PR #35, opened 2015-02-21
Use /bin/sh in the configure script

laurensa:
Not all systems come with Bash shell installed. Given that the configure script
is compatible with Bourne shell, /bin/bash can safely be replaced by /bin/sh.
This fixes the package configuration on systems without Bash or with Bash
installed elsewhere.

This commit to the configure script has been tested on a Bourne shell reimplementation. It fixes build on systems without Bash shell.

### DIFF
--- a/configure
+++ b/configure
@@ -1,16 +1,9 @@
-#!/bin/bash
+#!/bin/sh
 #
 # Copyright 2013 - 2014  Dirk Eddelbuettel
 #
 # Licensed under GPL 2 or later
 
-# This file uses bash.
-#
-# If this file is not suitable for your system for lack of bash or
-# another suitable /bin/sh implementation, I recommend deletion of the file
-# along with *manual* adjustment to RcppArmadilloLapack.h as per the
-# test below. 
-#
 # In most case you can just copy RcppArmadilloLapack.h.in over to
 # RcppArmadilloLapack.h -- and in aggregate it is not worth rewriting this.
 # At some point in the future we will be able to just rely on recent enough R


eddelbuettel:
That is too minor a pull request to be meaningful.  Also note _the text you chose to cut_ as well as the text that follows it:

```{.bash}
# This file uses bash.
#
# If this file is not suitable for your system for lack of bash or
# another suitable /bin/sh implementation, I recommend deletion of the file
# along with *manual* adjustment to RcppArmadilloLapack.h as per the
# test below. 
#
# In most case you can just copy RcppArmadilloLapack.h.in over to
# RcppArmadilloLapack.h -- and in aggregate it is not worth rewriting this.
# At some point in the future we will be able to just rely on recent enough R
# with good enough embeded Lapack, or use of external Lapack.
```

This really is not worth it.  It only concerns a subset of installations.

laurensa:
> That is too minor a pull request to be meaningful.

If you prefer other ways of addressing this problem, please let me know.

> Also note _the text you chose to cut_ as well as the text that follows it:

I cut the text since the configure script is working on Bourne shell. Bourne
shell is the original shell, Bash is not.

If you know the system has a /bin/sh shell, why not rely on that instead of
the much less common /bin/bash? If it turns out a user cannot get through
the configure script using /bin/sh on their system, let them submit patches
and assert the script is still working on the other systems for which it was
known to work before.

Also note that "deletion of the file along with *manual* adjustment" is not
possible in automated installs using `install.packages()`. Ideally `install.packages()`
should work on all systems for all packages. This commit tries to improve
the situation with respect to that.

> This really is not worth it.  It only concerns a subset of installations.

I am sorry, but you show complete ignorance about many Unix systems in use
today.

eddelbuettel:
>  It only concerns a subset of installations.

The issue at hand is a _substandard_ LAPACK/BLAS, and it is only those systems that need the file copy done by the shell script.   Do your homework before you attempt to make inferences over what I do or do know about Unix systems.

laurensa:
> The issue at hand is a substandard LAPACK/BLAS on some (older) installations
> as well as older R versions with their internal copies.  And it is only those
> systems that need the file copy done by the shell script.

Thank you for pointing it out. I see that only a subset of installations need
to actually make changes to RcppArmadilloLapack.h.in, but in order to figure
this out in an automatic fashion the script relies on Bash which is not
guaranteed to be installed on the system. Assuming Bash to be installed is what
I call ignorance. I also do not consider Bash to be a sane dependency, because
there are cross-platform ways to implement a shell script.

Automatic installation of this package on all systems without Bash currently
breaks and all affected users, even those that really do not care much about
this specific package, but had it listed as a dependency, must *manually*
follow the instructions in the configure script. I consider it a problem if the
package fails to pass automatic install. This is the problem I try to address
in this pull request and I also happen to be one of the affected users.

If you are not happy with my pull request but still agree with me that this is
a problem, please give me instructions how to change the pull request in a
manner such that it still fixes the problem I am seeing. If you on the other
hand disagree with me that this is a problem, then we can leave our discussion
here.

eddelbuettel:
The PR is closed and will remain closed.    The downside (to me as the one maintaining this) from changing from `bash` is *much larger* (been there, done that, had to change scripts) that the supposed pain inflicted on you for assuming you may have `/bin/bash`. 

If you system has room for `R`, it clearly has room for `bash`.  If you not, you get to alter the tarball manually. Your choice.  

This package is released to CRAN, and has a multi-year history of complying with the CRAN standards.

Issue #63, opened 2015-11-30
Support for Rcpp::as< arma::Cube<T> >


nathan-russell:
I took a shot at implementing a means of converting `Rcpp::Vector`s to `arma::Cube`s, but wanted to get some feedback on my code before submitting a PR because I think there are some minor sticking points worth discussing. The full change is reflected in [RcppArmadilloAs.h on my fork](https://github.com/nathan-russell/RcppArmadillo/blob/master/inst/include/RcppArmadilloAs.h#l121:l207), in the form of 

1. a specialization of `Rcpp::Exporter< arma::Cube<T> >`; and 
2. 3 further `Exporter` specializations, one for each of `arma::fcube`, `arma::ucube`, and `arma::cx_fcube` 

The first specialization works for fine for the `Cube<double>`, `Cube<sword>`, and `Cube<cx_double>`, but was choking on the remaining three typedefs. I could not come up with an elegant approach to encapsulate everything in one template, so I ended up with the additional three specializations that "piggy back" on their counterparts which *do* compile; e.g. to get an `arma::Cube<float>` / `arma::fcube`, a `SEXP` is first converted to an `arma::cube` using (1), and then converted to an `arma::fcube` using Armadillo's `conv_to< >::from` utility. Presumably this adds additional overhead which is not ideal, but as stated, I was unable to avoid this extra step. I am certainly open to any alternative suggestions however.  

In my [unit test file](https://github.com/nathan-russell/RcppArmadillo/blob/master/inst/unitTests/runit.cube.R) (using the C++ functions [here](https://github.com/nathan-russell/RcppArmadillo/blob/master/inst/unitTests/cpp/cube.cpp)), two of the tests involving `arma::cx_fcube`s use a modified tolerance value of `1.5e-7` when compared to base R's equivalent calculation. I suspect this has to do with a loss of precision in the conversion from `double` to `float`, but regardless I wanted to call attention to this.   

Finally, my `Exporter` specialization is written such that an exception will be thrown if the input object to be converted does not have *exactly* three dimensions. Is this desirable?  

Taking into account everything above, the corresponding Travis build was successful, but I certainly some discussion is in order.

eddelbuettel:
That is awesome. I meant to follow-up on the StackOverflow question suggesting exactly that.

Conrad is playing things old-school so we cannot ping him just by GitHub handle. I will send him an email.   

[...]

PR #64, opened 2015-11-30
Added support for conversion to arma::Cube types 

nathan-russell:
As discussed in #63 this extends `Rcpp::as<>` for use with Armadillo's templated `Cube` class.


eddelbuettel:
As expected, this passed well during the regression tests.  It is a really fine PR, and on reflection I think you need the extra specialization as they are all for cases where R does not have the underlying type: `float` or `unsigned`.  I also really like the explicit unit tests.

Issue #89, opened 2016-05-03
Possible memory leak with arma::sp_mat within Rcpp::List


gvegayon:
Hi, I don't know if this behavior is expected (I don't think so), but whenever I try to pass `dgCMatrix` objects within lists to `RcppArmadillo` via `arma::sp_mat` valgrind catches a memory leak. Here is a minimal example in which I create a list with dense matrices `arma::mat` and sparse matrices `arma::sp_mat`

The code for `arma::sp_mat`:

```r
library(Matrix)
library(Rcpp)

code <-
'// [[Rcpp::depends(RcppArmadillo)]]
#include <RcppArmadillo.h>
using namespace Rcpp;

// [[Rcpp::export]]
List passMats(List l) {
  arma::sp_mat m1 = as< arma::sp_mat >(l[0]);
  arma::sp_mat m2 = as< arma::sp_mat >(l[1]);

  return List::create(
    _["mat1"] = m1,
    _["mat2"] = m2
  );
}'

sourceCpp(code=code)

x <- matrix(0,4,4)
x <- methods::as(x, "dgCMatrix")
passMats(list(x,x))

devtools::session_info()
```
And this is the output that I get from running `R -d valgrind --vanilla`

```r
==3233== Memcheck, a memory error detector
==3233== Copyright (C) 2002-2013, and GNU GPL'd, by Julian Seward et al.
==3233== Using Valgrind-3.10.1 and LibVEX; rerun with -h for copyright info
==3233== Command: /usr/lib/R/bin/exec/R --vanilla
==3233== 

R version 3.2.5 (2016-04-14) -- "Very, Very Secure Dishes"
Copyright (C) 2016 The R Foundation for Statistical Computing
Platform: x86_64-pc-linux-gnu (64-bit)

R is free software and comes with ABSOLUTELY NO WARRANTY.
You are welcome to redistribute it under certain conditions.
Type 'license()' or 'licence()' for distribution details.

  Natural language support but running in an English locale

R is a collaborative project with many contributors.
Type 'contributors()' for more information and
'citation()' on how to cite R or R packages in publications.

Type 'demo()' for some demos, 'help()' for on-line help, or
'help.start()' for an HTML browser interface to help.
Type 'q()' to quit R.

> library(Matrix)
> library(Rcpp)
> 
> code <-
+ '// [[Rcpp::depends(RcppArmadillo)]]
+ #include <RcppArmadillo.h>
+ using namespace Rcpp;
+ 
[...]

PR #91, opened 2016-05-04
Fixing bug in SpMat exporter

gvegayon:
Solves issue https://github.com/RcppCore/RcppArmadillo/issues/89 as well as issue https://github.com/RcppCore/RcppArmadillo/issues/72. RcppArmadillo was compiled and checked with Valgrind and no memory leaks (or issues with `Conditional jump...`) where detected (nor in the examples, vignettes or tests). Also includes two new tests used to explicitly detect the two issues mentioned when using `R CMD check --use-valgrind`.


gvegayon:
Have no idea why it failed. The travis

PR #101, opened 2016-06-19
Rewrote sample.h to primarily use armadillo objects instead of rcpp objects

coatless:
http://stackoverflow.com/questions/37818872/stack-imbalance-while-using-rcppparallel/37819079#37819079


eddelbuettel:
Thanks for this.  I guess if the arma objects are created from R using the default approach they still involve the R-allocated memory.  I can't quite figure if that is, or is not, an issue.  Well, worst case folks can force deep copies...

PR #109, opened 2016-09-23
Force a deep copy in as<arma::Cube>(x) as in vec/mat cases

helske:
This PR tries to fix the issue discussed on the Rcpp-list, now a copy of x is always made when using as<>() for Cube, just like in case of vector and matrices.

The old and few new unit tests seems to pass fine, but to be honest I pretty much mimicked the codes from vec/mat case without deep understanding what I did, so... ;)

### DIFF
--- a/inst/include/RcppArmadilloAs.h
+++ b/inst/include/RcppArmadilloAs.h
@@ -117,8 +117,41 @@ namespace traits {
     // default Exporter-Cube specialization:
     // handles cube, icube, and cx_cube
     // fails on fcube, ucube, and cx_fcube
+    
+    // 23 September 2016
+    // Modified the Cube-exported to take account const ref
+    
+    //non-const ref case, make a copy
     template <typename T>
     class Exporter< arma::Cube<T> > {
+    public:
+        typedef arma::Cube<T> cube_t;
+        enum { RTYPE = Rcpp::traits::r_sexptype_traits<T>::rtype };
+        typedef typename Rcpp::traits::storage_type<RTYPE>::type value_t;
+        Exporter(SEXP x) : vec(x) {}
+      
+        cube_t get() {
+            Rcpp::Vector<INTSXP> dims = vec.attr("dim");
+            if (dims.size() != 3) {
+                std::string msg = 
+                  "Error converting object to arma::Cube<T>:\n"
+                  "Input array must have exactly 3 dimensions.\n";
+                Rcpp::stop(msg);
+            }
+          
+            cube_t result(
+                reinterpret_cast<T*>(vec.begin()),
+                dims[0], dims[1], dims[2], true);
+            return result;
+        }
+      
+    private:
+        Rcpp::Vector<RTYPE> vec;
+    };
+    
+    // const ref
+    template <typename T>
+    class Exporter< const arma::Cube<T>& > {
     public:
         typedef arma::Cube<T> cube_t;
         enum { RTYPE = Rcpp::traits::r_sexptype_traits<T>::rtype };
@@ -143,7 +176,6 @@ namespace traits {
     private:
         Rcpp::Vector<RTYPE> vec;
     };
-    
     // specializations for 3 cube typedefs that fail above
     // first use viable conversion SEXP -> Cube<other_t>
     // then use conv_to<cube_t>::from(other_t other)
@@ -276,6 +308,43 @@ namespace traits {
     
     /* End Armadillo vector as support classes */
     
+    template <typename T, typename CUBE, typename REF, 
+      typename NEEDS_CAST = typename Rcpp::traits::r_sexptype_needscast<T>::type>
+    class ArmaCube_InputParameter;
+    
+    template <typename T, typename CUBE, typename REF>
+    class ArmaCube_InputParameter<T, CUBE, REF, Rcpp::traits::false_type> {
+    public:
+      ArmaCube_InputParameter( SEXP x_ ) : c(x_), dims(c.attr("dim")), cube( reinterpret_cast<T*>( c.begin() ), 
+        dims[0], dims[1], dims[2], false ) {}
+      
+      inline operator REF(){
+        return cube ;        
+      }
+      
+    private:
+      Rcpp::Vector< Rcpp::traits::r_sexptype_traits<T>::rtype > c ;
+      Rcpp::Vector<INTSXP> dims;
+      CUBE cube ;
+      
+    } ;
+    
+    template <typename T, typename CUBE, typename REF>
+    class ArmaCube_InputParameter<T, CUBE, REF, Rcpp::traits::true_type> {
+    public:
+      ArmaCube_InputParameter( SEXP x_ ): c(x_), cube( as<CUBE>(c) ) {}
+      
+      inline operator REF(){
+        return cube ;        
+      }
+      
+    private:
+      Rcpp::Vector< Rcpp::traits::r_sexptype_traits<T>::rtype > c ;
+      CUBE cube ;
+    } ;
+    
+    /* End Armadillo vector as support classes */
+    
 #define MAKE_INPUT_PARAMETER(INPUT_TYPE,TYPE,REF)                       \
     template <typename T>                                               \
     class INPUT_TYPE<TYPE> : public ArmaVec_InputParameter<T, TYPE, REF >{ \
@@ -305,6 +374,19 @@ namespace traits {
     MAKE_INPUT_PARAMETER(ReferenceInputParameter     , arma::Mat<T>, arma::Mat<T>&       )
     MAKE_INPUT_PARAMETER(ConstInputParameter         , arma::Mat<T>, const arma::Mat<T>  )
 
+#undef MAKE_INPUT_PARAMETER
+
+#define MAKE_INPUT_PARAMETER(INPUT_TYPE,TYPE,REF)                             \
+    template <typename T>                                                   \
+    class INPUT_TYPE<TYPE> : public ArmaCube_InputParameter<T, TYPE, REF >{  \
+      public:                                                               \
+        INPUT_TYPE( SEXP x) : ArmaCube_InputParameter<T, TYPE, REF >(x){}    \
+    } ;                                                                                                                  
+    
+    MAKE_INPUT_PARAMETER(ConstReferenceInputParameter, arma::Cube<T>, const arma::Cube<T>& )
+    MAKE_INPUT_PARAMETER(ReferenceInputParameter     , arma::Cube<T>, arma::Cube<T>&       )
+    MAKE_INPUT_PARAMETER(ConstInputParameter         , arma::Cube<T>, const arma::Cube<T>  )
+      
 #undef MAKE_INPUT_PARAMETER
     
 }
--- a/inst/include/RcppArmadilloAs.h
+++ b/inst/include/RcppArmadilloAs.h
@@ -118,10 +118,7 @@ namespace traits {
     // handles cube, icube, and cx_cube
     // fails on fcube, ucube, and cx_fcube
     
-    // 23 September 2016
-    // Modified the Cube-exported to take account const ref
-    
-    //non-const ref case, make a copy
+    //make a copy when using as<arma::cube> like in vec/mat cases
     template <typename T>
     class Exporter< arma::Cube<T> > {
     public:
@@ -130,7 +127,7 @@ namespace traits {
         typedef typename Rcpp::traits::storage_type<RTYPE>::type value_t;
         Exporter(SEXP x) : vec(x) {}
       
-        cube_t get() {
+        cube_t get(bool copy = true) {
             Rcpp::Vector<INTSXP> dims = vec.attr("dim");
             if (dims.size() != 3) {
                 std::string msg = 
@@ -141,7 +138,7 @@ namespace traits {
           
             cube_t result(
                 reinterpret_cast<T*>(vec.begin()),
-                dims[0], dims[1], dims[2], true);
+                dims[0], dims[1], dims[2], copy);
             return result;
         }
       
@@ -149,33 +146,6 @@ namespace traits {
         Rcpp::Vector<RTYPE> vec;
     };
     
-    // const ref
-    template <typename T>
-    class Exporter< const arma::Cube<T>& > {
-    public:
-        typedef arma::Cube<T> cube_t;
-        enum { RTYPE = Rcpp::traits::r_sexptype_traits<T>::rtype };
-        typedef typename Rcpp::traits::storage_type<RTYPE>::type value_t;
-        Exporter(SEXP x) : vec(x) {}
-      
-        cube_t get() {
-            Rcpp::Vector<INTSXP> dims = vec.attr("dim");
-            if (dims.size() != 3) {
-                std::string msg = 
-                  "Error converting object to arma::Cube<T>:\n"
-                  "Input array must have exactly 3 dimensions.\n";
-                Rcpp::stop(msg);
-            }
-          
-            cube_t result(
-                reinterpret_cast<T*>(vec.begin()),
-                dims[0], dims[1], dims[2], false);
-            return result;
-        }
-      
-    private:
-        Rcpp::Vector<RTYPE> vec;
-    };
     // specializations for 3 cube typedefs that fail above
     // first use viable conversion SEXP -> Cube<other_t>
     // then use conv_to<cube_t>::from(other_t other)
@@ -185,7 +155,7 @@ namespace traits {
         typedef arma::fcube cube_t;
         
         Exporter(SEXP x)
-            : tmp(Exporter<arma::cube>(x).get()) {}
+            : tmp(Exporter<arma::cube>(x).get(false)) {}
         
         cube_t get() {
             cube_t result = arma::conv_to<cube_t>::from(tmp);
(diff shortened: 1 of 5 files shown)


eddelbuettel:
Thanks for that, and I need to think it through.  Part of me wants to change vec and mat to be like cube: no copies.  Not quite sure what is best -- will think about it.

helske:
Sure, I was thinking that too, you could always use `clone` if the copy is needed, and using it maybe makes the code more clear and less error-prone. In my own case I wanted to make a deep copy when picking components of a `List`, but someone else might want shallow copy, and `as<>` would probably be the easiest way for that.

I wonder how many users are currently relying on copy by `as<>` (perhaps unknowingly). Not sure if the current documentation actually says anything about which way it should work?

eddelbuettel:
Agreed on both points.

`as<>()` and `wrap()` were early. The very first interfaces used them to convert from/to a `SEXP`.  We then got better and lighter-weight with the proxies.  I have the feeling that `copy()` behaviour you saw was more or less unintended. Or maybe needed with more than two dimensions?

eddelbuettel:
Sorry that this feel to side -- my bad.  Trying to get back to this. 

One thing that gives me pause is that you say

> now a copy of `x` is always made when using `as<>()` for Cube, just like in case of vector and matrices.

what do you base that on?    When I look at RcppArmadilloAs.h I see only `false` as the argument to Armadillo's `copy_aux_mem`.  Where do you see copies being forced?

A lot of the rest of the PR is lovely and should go in.

inst/include/RcppArmadilloAs.h
```diff
@@ -125,7 +127,7 @@ namespace traits {
         typedef typename Rcpp::traits::storage_type<RTYPE>::type value_t;
         Exporter(SEXP x) : vec(x) {}
       
-        cube_t get() {
+        cube_t get(bool copy = true) {
```
eddelbuettel: Should this be `bool copy = false` to shadow the default value Armadillo use (and which you use below) ?

inst/include/RcppArmadilloAs.h
```diff
@@ -136,7 +138,7 @@ namespace traits {
           
             cube_t result(
                 reinterpret_cast<T*>(vec.begin()),
-                dims[0], dims[1], dims[2], false);
+                dims[0], dims[1], dims[2], copy);
```
eddelbuettel: With a default of `false` above, this would remain a proxy object copy.

helske:
Without this PR, I see this for `vec` (and `mat`):

```
# should alter the R object
cppFunction("void vectest1(arma::vec& x) { x(0) = 0.5; }", depends="RcppArmadillo")

# should not alter the R object
cppFunction("void vectest2(arma::vec x) { x(0) = 0.5; }", depends="RcppArmadillo")

# Maybe should, maybe not...
cppFunction("void vectest3(NumericVector x_) { 
  arma::vec x(as<arma::vec>(x_));
  x(0) = 0.5; }", depends="RcppArmadillo")

> y <- 1; vectest1(y); y
[1] 0.5
> y <- 1; vectest2(y); y
[1] 1
> y <- 1; vectest3(y); y
[1] 1
```

But for `cube`:
```
# should alter the R object
cppFunction("void cubetest1(arma::cube& x) { x(0,0,0) = 0.5; }", depends="RcppArmadillo")
# should not alter the R object
cppFunction("void cubetest2(arma::cube x) { x(0,0,0) = 0.5; }", depends="RcppArmadillo")
# maybe...
cppFunction("void cubetest3(NumericVector x_) { 
  arma::cube x(as<arma::cube>(x_)); 
  x(0,0,0) = 0.5; }", depends="RcppArmadillo")
> y <- array(1, c(1, 1, 1)); cubetest1(y); y
, , 1

     [,1]
[1,]  0.5

> y <- array(1, c(1, 1, 1)); cubetest2(y); y
, , 1

     [,1]
[1,]  0.5

> y <- array(1, c(1, 1, 1)); cubetest3(y); y
, , 1

     [,1]
[1,]  0.5

```

Especially the second case is clearly wrong right?

With this PR cube works like vec (last two function do not alter the object on R side), but like you said, not sure what would be the correct behaviour in `as<>()` case.

inst/include/RcppArmadilloAs.h
helske: That was the old behaviour, then we never get copies (see below). Only way I could circumvent this was to use copy as default here, but perhaps there is a more clever way.

eddelbuettel:
Thanks.  I am with you that the two cases are different.

But the question now is what they should be.  I'd argue that the cube case is preferable for all three cases.  In particular cases one and two (adding `&` or not) are somewhat spurious: we always have a `SEXP` underlying so the value will be altered:

```
R> library(Rcpp)
R> cppFunction("void foo1(NumericVector x) { x[0] = 0.5; }")
R> x <- 1; foo1(x); x
[1] 0.5
R> cppFunction("void foo2(NumericVector & x) { x[0] = 0.5; }")
R> x <- 1; foo2(x); x
[1] 0.5
R> x <- 1L; foo1(x); x     # forced copy
[1] 1
R> 
```

I think we should fix the Matrix/Vector case, rather than alter the Cube case which I find correct.

inst/include/RcppArmadilloAs.h
eddelbuettel: See below. I think this was the better behaviour and it should stay.

helske:
Hmm. Given your example with `NumericVector` it might be reasonable to have similar behaviour with arma objects as well, so yes maybe `cube` is correct and `vec/mat` are not. So I guess then the behaviour would be "always try to avoid copies". I personally do like the  current style of `vec/mat` though, which I feel is more consistent when comparing to regular C++. What I mean is that if I just see the function declaration like `void cubetest2(arma::cube x)`, I would expect that I can alter `x` without affecting the original object... Like in case of `void foo1(double x) { x = 0.5; }` (although I can't alter the original R object even with `double&`...). I regard `NumericVector` etc different as they are specific to `R/Rcpp` and thus one should know that they are just pointers to SEXP.

I would also guess that lots of user code gets broken if `vec/mat` if we stop making deep copies like in `cube` case. At least my codes ;)

conradsnicta:
I agree with helske.  For objects that are passed-by-value, making a shallow copy of them is inconsistent with the C++ standard.

Making shallow copies also causes a secondary inconsistency: a C++ function which is exposed to R in a user package will behave differently to a C++ function calling another C++ function within the same user package. This can cause all sorts of head-scratching bugs, and will be an endless source of confusion for newcomers.  In the end, C++ is not R, and pretending that it is only makes things worse.

A user can always specify that they explicitly want a shallow copy by specifying the argument as a reference, just like standard C++.

eddelbuettel:
@helske I just wanted to reiterate that I appreciate the report and issue ticket. I just want us to make the right choices here:
- we want to be consistent with the other parts of Rcpp (and there is no copy to value as all `SEXP` objects are s-expression-pointers) 
- so with that we may want to fix vec and mat to be more like cube and the rest, to be seen
- we also do not want to break existing code as you say (and I suspect most code returns as a return value rather than implicitly -- but we have to be very careful here)

@conradsnicta I can walk you through the details another time but in essence there CANNOT be a pass-by-value because of we talk back and forth to R and how the objects are represented (in an early 1990s C language design as C++ wasn't ripe enough then; the R team consists also almost exclusively of C programmers; such is life)

eddelbuettel:
Paging @nathan-russell :  As you did such fine work with Cube classes, any view from your side about whether / how to 'correct' mat and vec?

nathan-russell:
IIRC, when I wrote the code for `Exporter<arma::Cube>` I was primarily concerned with doing the conversion as efficiently as possible, hence the use of `copy_aux_mem = false` in the constructor; but admittedly hadn't considered whether or not this was consistent with the (non-reference semantic) behavior of `mat` and `vec`. 

As it stands, I can see the arguments on both sides: while it is important to provide a consistent API and not break users' existing code with new changes, I also firmly believe in not forcing people to pay for something they don't need -- namely, the default copying behavior or `mat` and `vec`. 

Regarding some of the earlier comments, while I realize that in "ordinary" C++ `function(T t)` implies that `t` passed by value, as Dirk pointed out, when working with `SEXP`-based classes we do not have all the luxuries of ordinary C++ semantics, which is why `function(Rcpp::NumericVector x)` is equivalent to `function(Rcpp::NumericVector& x)`. 
 
I have my hands full at work at the moment and can't do a proper assessment of whether it is possible to reimplement the `mat` and `vec` conversions *without* copying by default, but I would be more than happy to spend time on this over the weekend. If this change can be made with minimal impact on existing user code, it may very well be worthwhile. However, if the general consensus is that `Cube` should be modified to copy by default, then I can live with that as well.

eddelbuettel:
Thanks for so much for the follow-up, @nathan-russell .  This not urgent -- I made a mistake of forgetting about this issue til yesterday when I prepared the newest RcppArmadillo (currently at CRAN in incoming).

So we clearly have a few weeks to get this right, and we should give it a shot.  I'd also vote for 'consistent AND efficient' and would lean for a possible change.   E.g., for Rcpp I just made a (long overdue) change for better vectors of POSIXct object and protected that behind a `#define` we probably keep for a year.

helske:
I understand that cube case is more consistent with the other parts of Rcpp, although it feels inconsistent with C++ because armadillo objects are available outside of R/Rcpp sphere, whereas I have no problems with the behaviour of NumericVectors. So now function like `foo(arma::cube x)` behaves differently if it is exported to R, which feels weird to me. But if that is the common theme with other objects then I think change of vec/mat is the right way to go because of consistency (I would sacrifice default efficiency over safety in terms of side effects).

I too suspect that most code returns as a return value but still modify the passed arma objects at some intermediate point, and this change will cause lot's of headache in these situations as many unit tests etc. still work fine but when chained together...

eddelbuettel:
> Armadillo objects are available outside of R/Rcpp sphere

Sure. But we are having this very discussion about how to instantiate them from `SEXP` objects, and that is as R specific as it can be.  This is only about the interface to R.

For the rest we are all in agreement about preferring best C++ coding practices wherever possible.

nathan-russell:
After reading through the code in [RcppArmadilloAs.h](https://github.com/RcppCore/RcppArmadillo/blob/master/inst/include/RcppArmadilloAs.h) a bit more thoroughly, I realized that I had been overlooking the mechanics of [`MAKE_INPUT_PARAMETER`](https://github.com/RcppCore/Rcpp/blob/6f81b4684481dbd9bb554dd95e66725fc3b63a8c/inst/include/Rcpp/InputParameter.h#L27-L97) and how it is being used to enable `arma::whatever&` to treated appropriately as a reference type, and `arma::whatever` to be treated as a value type. I believe this is an oversight on my part, and had I noticed this when I originally wrote the `Cube` specialization for `Exporter`, I most likely would have added this in, as @helske [has done](https://github.com/helske/RcppArmadillo/blob/c56e7c76d6c4d96667b0085d0dcaa029ef15a385/inst/include/RcppArmadilloAs.h#L356-L367) in this PR. 

Incidentally, I had this realization while I was refactoring the `Col`, `Row`, and `Mat` specializations of `Exporter` to use Rcpp-style reference semantics (i.e. `arma::vec` would behave as `arma::vec&` currently does), so I now know for sure that defaulting to pass-by-reference across the board for `arma` types *is* an option. 


Sorry for being so equivocal about this, but I really could go either way at this point. 

-----------

As for 

> I too suspect that most code returns as a return value but still modify the passed arma objects at some intermediate point, and this change will cause lot's of headache in these situations as many unit tests etc. still work fine but when chained together... 

I had spent some time wondering whether a reverse dependency check would be an accurate gauge of client code breakage, and I think there's just really no way to know for sure. However, if we decide to go the route of pass-by-reference by default, introducing a deprecation period in the manner of Rcpp's improved datetime vector, and possibly adding a package startup message, should significantly help to mitigate any heartache.

eddelbuettel:
Thanks for the follow-up.  I stand corrected then.

Should we maybe jointly work on a branch and try to clean all this up?  I think there is no hurry--we definitely want to this right with the best balance between safety, efficiency and good standards.

nathan-russell:
Yes, that sounds like a plan to me.

helske:
Dirk, my comment was just to give a reasoning why I initially objected the `vec/mat` change and what actually caused me to stumble upon the whole problem; I routinely made sure not to alter the NumericVectors whereas I treated the passed arma objects differently.  But you have convinced me to think otherwise. 

I manually checked a random subset of functions in random packages linking to RcppArmadillo in github, and I think that the amount of errors caused by this change might not be as high as I initially thought, although automatic detection of those is a problem. Deprecation period and early notification to maintainers should help, and the needed changes to the code are pretty straightforward anyway.

nathan-russell:
I should also point out that in this scenario while 

```c++
void by_ref(arma::vec x) {
    x *= 2.0;
}
``` 

would modify the input vector, 

```c++
void by_val(arma::vec x) {
    arma::vec y = x;
    y *= 2.0;
}
``` 

would not, as what we do with `Exporter` has no bearing on `arma::whatever::operator=`. This is in contrast to 

```c++
void by_ref2(Rcpp::NumericVector x) {
    Rcpp::NumericVector y = x;
    y = y * 2.0;
}
``` 

which **does** modify its input. I think this is the primary "gotcha" for users who aren't aware of Rcpp's shallow copy semantics, and would be more likely to break code than what is being discussed here.

coatless:
Here's a question, how does `RcppEigen` and `RcppGSL` handle the interface to/from R? From what I recall about `RcppEigen`, it is pass-by-value unless specifically mapped via `Map`. My tour with `RcppGSL` is tour short to firmly recall the hand offs. However, I would strongly urge for consistency across all Rcpp core packages in the mannerism chosen for 3rd party C++ libraries, even if this means as Nathan put it, "forcing people to pay for something they don't need -- namely, the default copying behavior or mat and vec."

nathan-russell:
Off the top of my head I would guess that copies are made in those packages, although I can take a look see and find out for sure. The only reason we are able to get away with passing by reference at all  here is because these Armadillo objects all have constructor overloads which let you "steal" memory from an existing object.

helske:
It seems that everybody agree that consistency is the key here. I don't think the default behaviour needs to be the most efficient, but consistent and relatively safe, although the safety is more related to old users, as new users quickly learn what can be done and what not if things work consistently.

nathan-russell:
Hmm well it looks like RcppEigen *does* modify, 

```c++
// [[Rcpp::export]]
void eigen_ref(Eigen::Map<Eigen::MatrixXd> m) {
    m *= 2.0;
} 
```
but  RcppGSL does not: 

```c++
void gsl_ref(RcppGSL::Matrix m) {
    for (R_xlen_t j = 0; j < m.ncol(); j++) {
        for (R_xlen_t i = 0; i < m.nrow(); i++) {
            gsl_matrix_set(m, i, j, 9.0);
        }
    }
}
```

Unfortunately I'm not familiar with either of those frameworks so I can't really speak to how those *should* be handled.

eddelbuettel:
@coatless: RcppGSL is different and not relevant as they have a different underlying C represenation which need to get.  But it is also proxies.   RcppEigen _may_ be a model but to my mind that package is less polished that RcppArmadillo.   The good part of RcppArmadillo are really the best we got.

eddelbuettel:
@helsko I don't really agree with your

> It seems that everybody agree that consistency is the key here

As I said earlier, we balance three different goals: safety, efficiency, good standards.  

We can expand to consistency across packages once we have this first aggregate goal done right. RcppArmadillo is used by 274 other CRAN packages, RcppEigen only 72 (though lme4 among them, so a wide reach).  Basically only mvabund and RcppZiggurat use RcppGSL.  

So I would urge us to get RcppArmadillo right first, and then go from there.  As as shown in [this benchmark in the RcppArmadillo package around different ways to do fastLm](https://github.com/RcppCore/RcppArmadillo/blob/master/inst/examples/fastLm.r) we now have good performance (no copies) with good style ("looks like" references).

```r
edd@max:~$ Rscript git/rcpparmadillo/inst/examples/fastLm.r 
                       test replications relative elapsed
3         fLmConstRef(X, y)         5000    1.000   0.172
4             fLmSEXP(X, y)         5000    1.006   0.173
2         fLmTwoCasts(X, y)         5000    1.029   0.177
1          fLmOneCast(X, y)         5000    1.087   0.187
6   fastLmPureDotCall(X, y)         5000    1.198   0.206
5          fastLmPure(X, y)         5000    1.953   0.336
8              lm.fit(X, y)         5000    2.174   0.374
7 fastLm(frm, data = trees)         5000   28.134   4.839
9     lm(frm, data = trees)         5000   35.250   6.063
                     test replications relative elapsed
4       fLmConstRef(X, y)        50000    1.000   1.741
3           fLmSEXP(X, y)        50000    1.019   1.774
2       fLmTwoCasts(X, y)        50000    1.025   1.785
1        fLmOneCast(X, y)        50000    1.029   1.792
6 fastLmPureDotCall(X, y)        50000    1.182   2.058
5        fastLmPure(X, y)        50000    1.885   3.282
7            lm.fit(X, y)        50000    2.159   3.758
edd@max:~$ 
```

So: can we focus on RcppArmadillo, and figure out what if anything we want to change?   Shall we look into memory profiling to really be sure copies are made when we think they shouldn't?

coatless:
@eddelbuettel: Okay, point taken: RcppArmadillo is _the_ bright spot. Lessons learned will be distributed to other core packages.

eddelbuettel:
I am going to close this one.  

I think we have established that forcing a deep copy on Cube is actually not desirable.  I would strongly welcome a joint effort to establish what works well, what we like and what we would consider changing, as  discussed so far in this thread.  We could probably start that with a fresh issue ticket.   I think it would also be a good idea to do some experiments related to timing (maybe expanding / reworking the old and simple `fastLm()` benchmark) in terms of cpu profiling / benchmarking, but also memory profiling.

PR #130, opened 2017-05-10
Clarified RNG usage in C++98 and C++11 modes. (close #117)

### LINKED ISSUE #117: Clarify documentation (in Rd files) on RNG use
Tickled by [this SO question](http://stackoverflow.com/questions/41250974/what-is-the-random-number-generator-that-armadillo-uses).

coatless:


R/RcppExports.R
```diff
@@ -19,11 +19,13 @@ armadillo_version <- function(single) {
...
 #' Set the Armadillo Random Number Generator to a random value
 #' 
-#' @details Depending on whether RcppArmadillo was compiled for the C++98 standard 
-#' (currently the default) or for C++11 (optional), two different RNGs may be used. 
-#' This function resets either. For C++98, the system library RNG is used. It may
-#' have poor performance, particularly on older versions of Windows. For C++11, the
-#' RNG included in its library is used.
+#' @details
+#' Depending on whether RcppArmadillo was compiled for the C++98 standard
+#' (currently the default) or for C++11 (optional), two different RNGs may be used.
```
eddelbuettel: That is about to change with 7.900.0 with C++11 the default.

eddelbuettel: Yes, get some sleep :)

The plan is to hardwire this in `src/Makevars`.  But come to think about it, that does not bubble over to users automagically.  So we also have to update the plugins and examples.  

Not a small change.  

I cherry-picked your PR over to the 7.900.0.  We can have a four eyes on that one.

PR #137, opened 2017-06-19
update license for compatibility

conradsnicta:
Apache 2.0 license is compatible with GPL 3.
https://en.wikipedia.org/wiki/Apache_License#GPL_compatibility
"... The Free Software Foundation considers all versions of the Apache License to be incompatible with the previous GPL versions 1 and 2 ..."


eddelbuettel:
Closing this as you for some reason filed an identical one minutes later.

PR #138, opened 2017-06-19
update license for compatibility

conradsnicta:
Apache 2.0 license is compatible with GPL 3.
https://en.wikipedia.org/wiki/Apache_License#GPL_compatibility
"... The Free Software Foundation considers all versions of the Apache License to be incompatible with the previous GPL versions 1 and 2 ..."

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -19,7 +19,7 @@ Description: 'Armadillo' is a templated C++ linear algebra library (by Conrad
  the GNU GPL version 2 or later, as is the rest of 'Rcpp'. Note that
  Armadillo requires a fairly recent compiler; for the g++ family at least
  version 4.6.* is required. 
-License: GPL (>= 2)
+License: GPL (>= 3)
 LazyLoad: yes
 LinkingTo: Rcpp
 Imports: Rcpp (>= 0.11.0), stats, utils


eddelbuettel:
I fear you may be misunderstanding something.  I make no statement about your license.

The DESCRIPTION file pertains to the R package, and the code it adds.  That code (of mine and others) is under the GPL (>= 2).  

It so happens that the tarball also includes files of yours. About which I say the following [in lines 13 to 21](https://github.com/RcppCore/RcppArmadillo/blob/master/DESCRIPTION#L13-L21)

>  The 'RcppArmadillo' package includes the header files from the templated
 'Armadillo' library. Thus users do not need to install 'Armadillo' itself in
 order to use 'RcppArmadillo'. From release 7.800.0 on, 'Armadillo' is
 licensed under Apache License 2; previous releases were under licensed as
 MPL 2.0 from version 3.800.0 onwards and LGPL-3 prior to that;
 'RcppArmadillo' (the 'Rcpp' bindings/bridge to Armadillo) is licensed under
 the GNU GPL version 2 or later, as is the rest of 'Rcpp'. Note that
 Armadillo requires a fairly recent compiler; for the g++ family at least
 version 4.6.* is required. 

If there is anything you feel that needs to be added?  If not I plan to close this.

conradsnicta:
Well, mixing GPL v2 licensed code with Apache 2.0 licensed code doesn't work.
RcppArmadillo does extend the Mat + etc classes in Armadillo, meaning it is mixing code.

eddelbuettel:
For that you have GPL-3 as "GPL (>= 2)" gives you a choice.

eddelbuettel:
Also see https://github.com/conradsnicta/armadillo-code/blob/unstable/LICENSE.txt which, when viewed at GitHub, gets you this fancy license viewer illustrating terms which shows in particular

> Licensed works, modifications, and larger works may be distributed under different terms and without source code.

which seems explicitly permit what we're doing here.  Nobody infringes on your terms or copyright.  We simply use a different one for out work.  

Which we haven't changed while you have gone through three different licenses.

Issue #114, opened 2016-11-14
RcppArmadillo sp_mat doesn't interpret Sparse Symmetric Matrix correctly


jeffwong-nflx:
This code produces a sparse symmetric matrix, a special case when crossprod(x) is known to be symmetric.

```
require(Matrix)
x = rsparsematrix(10000, 50, .05)
x2 = crossprod(x)
```
Note that the nonzero values stored in x2 only reflect the upper triangle

```
x2@x #only upper triangle part!
```

When this is sent to RcppArmadillo as a sp_mat, only the upper triangle is sent, and it thinks that the lower triangle is all 0

```
#include <RcppArmadillo.h>
// [[Rcpp::depends(RcppArmadillo)]]

using namespace Rcpp;
using namespace arma;

// [[Rcpp::export]]
sp_mat foo(sp_mat X) {
  return X;
}
foo(crossprod(rsparsematrix(10000, 5, .01)))
```

eddelbuettel:
Thanks for the bug report, and for bringing it here.  I will try to take a look.

I think the overarching issue may be that we have a larger number of possibly representations in the Matrix package than are currently handled by the converters.  So it will require someone to sit down and write those.

thirdwing:
This seems very similar to #17 .

I will take a look in the weekend.

thirdwing:
I was thinking using `symmatu() / symmatl()` to handle the symmetric matrix. However, it only works on `arma::mat` not `arma::sp_mat`.

More complicated than I thought.

coatless:
FYI: 

Sparse support was just added today for...

- `symmatu()` / `symmatl()` in https://github.com/conradsnicta/armadillo-code/commit/d99b8241e299e61058038294001e71acb77e7f74
- `trimatu()`/`trimatl()` in  https://github.com/conradsnicta/armadillo-code/commit/f294d3bca5508cc44ce2a744f6b908854372794b

thirdwing:
WOW~!

Let's complete GSOC project idea!

eddelbuettel:
@conradsnicta being Conrad, I am of course prohibited from following the repo.  #facepalm

coatless:
You can prohibit someone from following a repo?

eddelbuettel:
Apparently.  Looks like you snug in before he changes that setting.

jeffwong-nflx:
[...]

Issue #17, opened 2014-06-04
implement for as<arma::sp_mat> for Matrix classes other than dgCMatrix


fabian-s:
It would nice if RcppArmadillo knew how to deal with all Matrix-objects inheriting from `CsparseMatrix` (i.e. all Matrix classes in sorted compressed column-oriented form), not just `dgCMatrix`.  The `ngCMatrix`-class doesn't have an x-slot, for example, because the entries are all TRUE/FALSE, so the current implementation of `as<arma::sp_mat>` fails. 
Of course it's usually possible  to do a conversion to `dgCMatrix` on the R-side, but that's messy -- right now, in R you have to do something like 

`as(as(X, "CsparseMatrix"), "dMatrix")`

to every Matrix-object you're handing over to RcppArmadillo to be sort-of-safe that it ends up having the correct class -- but the above will still fail to create a `dgC`-matrix if X inherits from `symmetricMatrix`, for example. You have to do the conversion in this roundabout way because, as Martin Maechler explicitly writes in the Matrix docs,  `as(X, <MatrixClass>)` should only be used to convert `X`  to "virtual Matrix super classes" like `CsparseMatrix` and not actual classes  like `dgCMatrix`, and there aren't any direct `as`-methods to make a `dgCMatrix` from, say, an `ngTMatrix` or a `ddiMatrix`.

These conversions in R can be costly as the matrix gets copied around, it would be more efficient to create the `sp_mat` directly. If that is too difficult, the RcppArmadillo-converter could at least check the class of its input to give more informative/correct error messages. 

Thanks for all your great work on these packages!

thirdwing:
Believe it or not, I am a little surprised about how many matrix classes defined in the `Matrix` package (https://github.com/cran/Matrix/blob/master/src/Mutils.h#L349).

Personally I think at least we need to provide some warning message if it is not `dgCMatrix`.

eddelbuettel:
Agreed on the idea for a warning matrix.  

Sparse support simple is better in (Rcpp)Eigen due to the needs by Doug Bates at the time. Until someone _really_ cranks sparse data with (Rcpp)Armadillo this will remain limited.  A related problem is, of course, the lack of actual sparse matrix transformations as you discovered yesterday in #114.

eddelbuettel:
I guess that is the second time today that I am thinking "hm, maybe for next year's Google Summer of Code".  Someone remind me when the time is closer :)

thirdwing:
[...]

PR #139, opened 2017-06-19
Add conversion & unit tests for dgR, dtR & dsRMatrix

binxiangni:
In order to solve #17 and #114. And there is a piece of codes using the algorithm from [scipy](https://github.com/scipy/scipy/blob/master/scipy/sparse/sparsetools/csr.h#L436). I have added a reference to that. What else should I do to avoid violating the copy rights? @coatless @eddelbuettel @thirdwing 

Btw, I squash and merge a branch called Rsparse to the master, but now the commit history still shows up here. I know the reason might be that I didn't squash and merge the previous commits, which leads to a messy timeline. :(

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,8 +1,18 @@
+
+2017-06-19  Binxiang Ni  <binxiangni@gmail.com>
+
+	* inst/include/RcppArmadilloAs.h: Add conversion for dgR, dtR & dsRMatrix
+	* inst/unitTests/runit.sparse.R: Add unit tests for the conversion for dgR, dtR & dsRMatrix
+
+2017-06-18  Binxiang Ni  <binxiangni@gmail.com>
+
+	* inst/include/RcppArmadilloAs.h: Add conversion for dgT, dtT & dsTMatrix
+	* inst/unitTests/runit.sparse.R: Add unit tests for the conversion for dgT, dtT & dsTMatrix
+
 2017-06-17  Binxiang Ni  <binxiangni@gmail.com>
 
-        * inst/include/RcppArmadilloAs.h: Add conversion for dtCMatrix & dsCMatrix
-        * inst/unitTests/runit.sparse.R: Add unit tests for the conversion for      
-        dtCMatrix & dsCMatrix
+	* inst/include/RcppArmadilloAs.h: Add conversion for dtC & dsCMatrix
+	* inst/unitTests/runit.sparse.R: Add unit tests for the conversion for dtC & dsCMatrix
 
 2017-06-04  Dirk Eddelbuettel  <edd@debian.org>
 
--- a/inst/include/RcppArmadilloAs.h
+++ b/inst/include/RcppArmadilloAs.h
@@ -167,7 +167,10 @@ namespace traits {
                     p[col] = cumsum;
                 }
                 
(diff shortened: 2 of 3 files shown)


eddelbuettel:
I think you are doing something wrong.

You have not pulled or merged or ... the master branch.  Most of these commits here we already accounted for.

binxiangni:
@eddelbuettel Perhaps I should sync the newest RcppCore/RcppArmadillo to my repo binxiangni/RcppArmadillo first and then commit on that?

eddelbuettel:
It is worth trying one or two things with git.  You can rename your current RcppArmadillo working directory and copy of the repo -- say, `RcppArmadillo-previous` (or add the date, or ...).   

The create a fresh checkout.   That corresponds to our master branch here.

Then try to get both aligned.   You should be able to the the 'working' copy (ie 'previous' above) to just about the same set of commits -- differing only by your _new_ commits made since I merged the branch.

coatless:
Next time around, run `git pull --rebase <remote-name> <branch-name>` before starting on a PR. This should bring your branch up to the present base without a "Merge branch 'master' of github.com:user/repo"

See [Atlassian's Git Rebasing](https://www.atlassian.com/git/tutorials/rewriting-history) for more.

eddelbuettel:
Why rebase and not just merge?  The language about 'new commits' tends to confuse me because I do want the existing commits, not the old changes recommitted as new ones. 

I think what I do in such a case is to
- pull the remote master (ie this repository) to be current
- branch cleanly ie with a fresh `git checkout -b feature/newstuff`  which then contains all of master

But this is git so there are probably half a dozen ways...

eddelbuettel:
That looks much better :100:

eddelbuettel:
Looks really good.  Two quick questions:

1) You mentioned
> And there is a piece of codes using the algorithm from scipy.
Where is that? Does it need a attribution in the file header (i.e. something like 'Portions taken from file abc.py written by D, E and F and released under copyright G' ?)

2) So what did you do for the cleaner PR?  Just a pull from our master, and push?   It did the right thing as all the commits we already accounted

binxiangni:
1. 

You can look at the reference [here](https://github.com/binxiangni/RcppArmadillo/blob/master/inst/include/RcppArmadilloAs.h#L265). If attribution can be added, that would be great. It can be "Portions taken from file [csr.h](https://github.com/scipy/scipy/blob/master/scipy/sparse/sparsetools/csr.h#L436) written by SciPy developers and released under Copyright 2017 SciPy developers. "(?) I only find that on their website.

2. 

- I git clone binxiangni/RcppArmadillo to local and reset to the commit without all my previous commits. 

- pull from RcppCore/RcppArmadillo

- paste the three modified files to the local repo, git add and git commit

- git push -f to binxiangni/RcppArmadillo

eddelbuettel:
1. My bad. I missed the reference.  Yes. something like that would do.  I may be hard to get concise author reference for SciPy.  

2. I see. I didn't account for  the `git push -f` possibility.  

All good.

eddelbuettel:
All merged.  

@binxiangni can you maybe write a short note at each of e #17 and #114 and close them, if appropriate, now that this is merged?

PR #140, opened 2017-06-20
Add conversion & unit tests for pMatrix & ddiMatrix

binxiangni:
In order to solve #17 and #114 . Now the conversion for all the numeric sparse matrices is done. @thirdwing @eddelbuettel When doing unit tests, I found some bugs in the package Matrix. But there is only a read-only repo on the GitHub so that I cannot open a pull request. Do you have any idea where I can contribute?

### DIFF
--- a/inst/include/RcppArmadilloAs.h
+++ b/inst/include/RcppArmadilloAs.h
@@ -79,7 +79,7 @@ namespace traits {
     };
     
     // 14 June 2017
-    // Add support for dgCMatrix, dtCMatrix and dsCMatrix
+    // Add support for sparse matrices other than dgCMatrix
     template <typename T>
     class Exporter< arma::SpMat<T> > {
     public:
@@ -249,7 +249,7 @@ namespace traits {
                 int nnz = rx.size();
                 IntegerVector i = IntegerVector(nnz);
                 IntegerVector p = IntegerVector(ncol + 1);
-                NumericVector x = NumericVector(nnz);
+                Vector<RTYPE> x = Vector<RTYPE>(nnz);
               
                 // Count the nnz in each column
                 for(int n = 0; n < nnz; n++){
@@ -300,7 +300,7 @@ namespace traits {
                 int nnz = rx.size();
                 IntegerVector i = IntegerVector(nnz);
                 IntegerVector p = IntegerVector(ncol + 1);
-                NumericVector x = NumericVector(nnz);
+                Vector<RTYPE> x = Vector<RTYPE>(nnz);
                 
                 // Count the nnz in each column
                 for(int n = 0; n < nnz; n++){
@@ -355,7 +355,7 @@ namespace traits {
                 int nnz = rx.size();
                 IntegerVector i = IntegerVector(nnz);
                 IntegerVector p = IntegerVector(ncol + 1);
-                NumericVector x = NumericVector(nnz);
+                Vector<RTYPE> x = Vector<RTYPE>(nnz);
                 
                 // Count the nnz in each column
                 for(int n = 0; n < nnz; n++){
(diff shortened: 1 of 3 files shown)


eddelbuettel:
I _think_ Matrix lives [here as an SVN repo on R-Forge](http://r-forge.r-project.org/projects/matrix/). That is consistent with its DESCRIPTION file.

inst/include/RcppArmadilloAs.h
```diff
@@ -403,14 +403,61 @@ namespace traits {
                     res = symmatl(res);
                 }
             }
+            else if (type == "pMatrix") {
+                IntegerVector i(ncol);
+                IntegerVector p(ncol+1);
```
coatless: Not sure about this one...

Consider: 

```r
(pm1 <- as(as.integer(c(2,3,1)), "pMatrix"))
# 3 x 3 sparse Matrix of class "pMatrix"
# 
# [1,] . | .
# [2,] . . |
#   [3,] | . .
> str(pm1)
# Formal class 'pMatrix' [package "Matrix"] with 4 slots
# ..@ perm    : int [1:3] 2 3 1
# ..@ Dim     : int [1:2] 3 3
# ..@ Dimnames:List of 2
# .. ..$ : NULL
# .. ..$ : NULL
# ..@ factors : list()
```

binxiangni: Sorry, what do you mean? 

For the example you provide, when converted to dgCMatrix,
`i: 0 1 2
 p: 0 1 2 3
 x:  1 1 1`

binxiangni: I perhaps know what you mean. Let me see...

binxiangni: This points to a big problem I never thought of. 
When 
1. nrow = ncol;
2. all the non-zero values are the same;
3. each row and each column only has a non-zero values

The converted dgCMatrix will not carry all the information of the matrix, which means the further mathematic operation based on the converted dgCMatrix might be meaningless.

binxiangni: For now, I think the pMatrix might not be converted to dgCMatrix since different pMatrix can lead to the same dgCMatrix, but if you have any ideas, please let me know. I'll add warning messages in the conversion of pMatrix.

thirdwing: Also not sure about converting permutation matrix into a numeric matrix.

Maybe we can discuss this in another issue or PR. How about removing this for now?

binxiangni: Wait, I think the same problem might also exist in ddiMatrix. I am thinking of another solution.

binxiangni: Now the bug for pMatrix has been fixed. I used the matrix James provided and passed the unit test locally.

inst/include/RcppArmadilloAs.h
```diff
@@ -403,14 +403,68 @@ namespace traits {
...
+                
+                // Sort the row number by the column number
+                std::map <int, int> colrow;
+                for(int tmp = 0; tmp < perm.size(); tmp++){
+                    colrow[perm[tmp]] = tmp;
+                }
+                
+                // Calculate i
+                for(std::map<int, int>::iterator tmp = colrow.begin(); tmp != colrow.end(); tmp++){
+                    i.push_back(tmp -> second);
```
coatless: Using `.push_back` is problematic on `IntegerVector` because Rcpp are proxy models that serve as shallow wrappers around R objects. Thus, each time you add an element, there is a significant overhead associated with copying the existing vector to new vector. It would be much better in this case to directory use `std::vector<int>` and then cast at the end to `IntegerVector`.

inst/include/RcppArmadilloAs.h
```diff
@@ -403,14 +403,68 @@ namespace traits {
...
+                // Making space for the elements
+                res.mem_resize(static_cast<unsigned>(x.size()));
+              
+                // Copying elements
+                std::copy(i.begin(), i.end(), arma::access::rwp(res.row_indices));
+                std::copy(p.begin(), p.end(), arma::access::rwp(res.col_ptrs));
+                std::copy(x.begin(), x.end(), arma::access::rwp(res.values));
+            }
+            else if (type == "ddiMatrix") {
+                IntegerVector i(ncol);
```
coatless: This also needs to be tweaked slightly as it does not take into account `N` or `U`  classes of the [`ddiMatrix`](https://stat.ethz.ch/R-manual/R-devel/library/Matrix/html/ddiMatrix-class.html).


e.g.

```r
(d_U <- Diagonal(n = 3))
# 3 x 3 diagonal matrix of class "ddiMatrix"
#     [,1] [,2] [,3]
# [1,]    1    .    .
# [2,]    .    1    .
# [3,]    .    .    1
str(d_U)
# Formal class 'ddiMatrix' [package "Matrix"] with 4 slots
#  ..@ diag    : chr "U"
#  ..@ Dim     : int [1:2] 3 3
#  ..@ Dimnames:List of 2
#  .. ..$ : NULL
#  .. ..$ : NULL
#  ..@ x       : num(0) 
```
```
(d_N <- Diagonal(x = c(10,1)))
# 2 x 2 diagonal matrix of class "ddiMatrix"
#      [,1] [,2]
# [1,]   10    .
# [2,]    .    1
str(d_N)
# Formal class 'ddiMatrix' [package "Matrix"] with 4 slots
#  ..@ diag    : chr "N"
#  ..@ Dim     : int [1:2] 2 2
#  ..@ Dimnames:List of 2
#  .. ..$ : NULL
#  .. ..$ : NULL
#  ..@ x       : num [1:2] 10 1
```

eddelbuettel:
I think I can merge this now.  Any objections?

coatless:
Sorry for not seeing this earlier. 

Everything looks to be in a better state.

PR #142, opened 2017-07-05
Rewrite conversion for dgT, dtT & dsTMatrix & add file-based tests

binxiangni:
First, the file-based tests are based on the documentation from package Matrix, sparseM, spam, slam, SciPy.   

Second, when I did the file-based tests, I found two extreme cases that the original conversion for TsparseMatrix cannot satisfy, so I rewrite the conversion.

The two new cases include:
1. [i is increasingly ordered](https://github.com/binxiangni/RcppArmadillo/blob/master/tests/testSparse.R#L183);
2. [(i, j) pair is repeated](https://github.com/binxiangni/RcppArmadillo/blob/master/tests/testSparse.R#L46).

Last, I've passed the tests, but there is a warning that 
`Warning in read_symbols_from_dll(so, rarch) :`
    `this requires 'objdump.exe' to be on the PATH`
Should I do something?

### DIFF
--- a/inst/unitTests/runit.sparseConversion.R
+++ b/inst/unitTests/runit.sparseConversion.R
@@ -1,0 +1,676 @@
+#!/usr/bin/r -t
+#
+# Copyright (C) 2014  Dirk Eddelbuettel
+#
+# This file is part of RcppArmadillo. It is based on the documentation of package Matrix, 
+# slam, SparseM, spam and SciPy, which are respectively created by developers of the packages:
+# Douglas Bates, Martin Maechler; Kurt Hornik, David Meyer, Christian Buchta; 
+# Roger Koenker, Pin Ng; Reinhard Furrer, Florian Gerber, Daniel Gerber, 
+# Kaspar Moesinger, Youcef Saad, Esmond G. Ng, Barry W. Peyton, Joseph W.H. Liu,
+# Alan D. George; the developers of SciPy. It is also modified by Binxiang Ni.
+#
+# RcppArmadillo is free software: you can redistribute it and/or modify it
+# under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 2 of the License, or
+# (at your option) any later version.
+#
+# RcppArmadillo is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with RcppArmadillo.  If not, see <http://www.gnu.org/licenses/>.
+
+# Reference:
+# [Matrix]: https://cran.r-project.org/web/packages/Matrix/Matrix.pdf
+# [slam]: https://cran.r-project.org/web/packages/slam/slam.pdf
+# [SparseM]: https://cran.r-project.org/web/packages/SparseM/SparseM.pdf
+# [spam]: https://cran.r-project.org/web/packages/spam/spam.pdf
+# [SciPy]: https://docs.scipy.org/doc/scipy/reference/sparse.html
+
+.runThisTest <- suppressMessages(require(Matrix))
+
+if (.runThisTest) {
+  
+    .setUp <- RcppArmadillo:::unit_test_setup("sparse.cpp") 
+  
+    ## setting up an example matrix -- using the fact that the as<sp_mat>
+    ## converter prefers sparse matrix objects create by the Matrix package
+    suppressMessages(require(Matrix))
+    
+    test.as.dgc2dgc <- function() {
+        ## [Matrix] p10 (dgCMatrix)
+        set.seed(7)
+        m <- matrix(0, 5, 5)
+        m[sample(length(m), size = 14)] <- rep(1:9, length=14)
+        mm <- as(m, "CsparseMatrix")
+        checkEquals(mm, asSpMat(mm), msg="dgC2dgC_1")
+        
+        ## [Matrix] p36 (dgCMatrix)
+        m <- Matrix(c(0,0,2:0), 3,5)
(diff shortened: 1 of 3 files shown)


eddelbuettel:
The change to the header file looks good and careful -- though I only glanced at it.  It is quite possible that you fixed an error which had been hiding there for some time.  

I am wondering if we can do better on the (short) included cpp file?   Ie for the rest of the package we have the cpp files [in this directory](https://github.com/RcppCore/RcppArmadillo/tree/master/inst/unitTests/cpp) which already has a file for [sparse matrix tests](https://github.com/RcppCore/RcppArmadillo/tree/master/inst/unitTests/cpp).  Could you use that mechanism, and call via unit test functions as in [this file](https://github.com/RcppCore/RcppArmadillo/blob/master/inst/unitTests/runit.sparse.R)  ?

binxiangni:
Maybe I don't totally understand what you mean, but for now,I just call the [cpp file](https://github.com/RcppCore/RcppArmadillo/blob/master/inst/unitTests/cpp/sparse.cpp#L26) to do the unit tests. The unit test is so long just because I have to write a new triangular and a new symmetric matrix to test them.

eddelbuettel:
You added two files to subdirectory `tests/`.  That is useful for the simplest test cases.   But yours is already more complex. 

And what the package does internally is to use the RUnit package to have the file [tests/doRUnit.R](https://github.com/RcppCore/RcppArmadillo/blob/master/tests/doRUnit.R) automatically call everything in the [unit tests directory](https://github.com/RcppCore/RcppArmadillo/tree/master/inst/unitTests).  

And we _already_ have 1) a file for sparse tests there which you could add two, or complement witha  new one, and 2) a C++ file there to support it -- so you short C++ function could go there.

Have a look at those files, and what happens when you do `R CMD check RcppArmadillo_*tar.gz`.   It would be better to have your new tests there.

eddelbuettel:
Also, what we could do is too 
- first merge as is
- then post-process to convert into the standard format
I could do that if you want me to. And I'll make sure you end up in the corresponding copyright headers.

binxiangni:
I've integrated the previous files into the [unitTests](https://github.com/binxiangni/RcppArmadillo/blob/master/inst/unitTests/runit.sparseConversion.R) folder. Please help me check if the format and copyright headers are appropriate. 

And I passed the tests locally, but it's weird it didn't work on the website. I checked the log and the error happened [here](https://travis-ci.org/RcppCore/RcppArmadillo/builds/250494091#L999). But I didn't touch the [runit.sparse.R](https://github.com/binxiangni/RcppArmadillo/blob/master/inst/unitTests/runit.sparse.R).

thirdwing:
It seems the unit test also passed on my machine.

Let't restart the Travis and see what happens.

inst/unitTests/runit.sparseConversion.R
```diff
@@ -0,0 +1,593 @@
...
+    .setUp <- RcppArmadillo:::unit_test_setup("sparse.cpp") 
+  
+    ## setting up an example matrix -- using the fact that the as<sp_mat>
+    ## converter prefers sparse matrix objects create by the Matrix package
+    suppressMessages(require(Matrix))
+
+    # Matrix
+    # https://cran.r-project.org/web/packages/Matrix/Matrix.pdf
+    
+    ## p10 (dgCMatrix)
```
thirdwing: I think you mean page 10, right? Just want to make sure.

eddelbuettel:
Thanks for making the chamges!  I can take a look tomorrow.  Bed-time here in Brussels...

inst/unitTests/runit.sparseConversion.R
```diff
@@ -0,0 +1,593 @@
...
+              "0 20 0 0 0",
+              "0 0 30 0 0",
+              "0 0 0 40 0",
+              "0 0 0 0 50")
+    M <- as.matrix(read.table(textConnection(mtxt)))
+    dimnames(M) <- NULL
+    dgc <- as(M, "dgCMatrix")
+    checkEquals(dgc, asSpMat(D5), msg="ddi2dgC")
+    
+    # mtxt <- c("0.1 0    0          0     0",
```
thirdwing: Can I know why this test has been commented out?

inst/unitTests/runit.sparseConversion.R
binxiangni: Yes, it represents the 10th page of documentation of package Matrix.

inst/unitTests/runit.sparseConversion.R
binxiangni: This test didn't pass. I guess it's because the rounding precision. I leave it for further check.

thirdwing:
@binxiangni I am afraid you are using a wrong format for RUnit.

For example, in https://github.com/binxiangni/RcppArmadillo/blob/master/inst/unitTests/runit.sparseConversion.R#L39-L43

```r
set.seed(7)
m <- matrix(0, 5, 5)
m[sample(length(m), size = 14)] <- rep(1:9, length=14)
mm <- as(m, "CsparseMatrix")
checkEquals(mm, asSpMat(mm), msg="dgC2dgC")
```

This should be 
```r
    test.as.dgC2dgC <- function() {
        set.seed(7)
        m <- matrix(0, 5, 5)
        m[sample(length(m), size = 14)] <- rep(1:9, length=14)
        mm <- as(m, "CsparseMatrix")
        checkEquals(mm, asSpMat(mm), msg="dgC2dgC")
    }
```

This should make Travis CI happy. See https://travis-ci.org/thirdwing/RcppArmadillo

binxiangni:
Now the new unit test file should be fine. This is a good lesson for me to learn how to do unit tests in RUnit. Thanks for instruction. :)

eddelbuettel:
Yes, looks much better :)

With this in line 552:

> # ....(To be continued)

Do you want to complete this, or shall we merge now?

binxiangni:
That's for the next stage, since some of conversion is not achieved now. I just note that in order to remind myself. You can merge that, if no error is found.

PR #144, opened 2017-07-08
Add conversion & unit tests for indMatrix

binxiangni:
This is the last numeric sparse matrix left. So far the conversion for the numeric sparse matrix should be done.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,11 @@
+2017-07-08  Binxiang Ni  <binxiangni@gmail.com>
+
+	* inst/include/RcppArmadilloAs.h: Add conversion for indMatrix
+	* inst/unitTests/runit.sparse.R: Add unit tests for the conversion
+	for indMatrix
+	* inst/unitTests/runit.sparseConversion.R: Add more unit tests for 
+	the conversion for indMatrix
+
 2017-07-08  Dirk Eddelbuettel  <edd@debian.org>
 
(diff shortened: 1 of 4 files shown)


eddelbuettel:
You did really well with the previous PRs!  Just check the one commit I did earlier.  In general when you use `textConnection()` you also need to close the connection.  So I switched to the newer `text=...` argument.  No other changes from my end.

Oh, and just one thing on wording:  you did not implement "file-based" tests as we do not use files :)  You could say text-based, or something.   Doesn't really matter.

binxiangni:
Just finished switching to `text=...`. There are some tests for sparse matrix conversion in [runit.sparse.R](https://github.com/binxiangni/RcppArmadillo/blob/master/inst/unitTests/runit.sparse.R#L95-L257), should I move them tor [runit.sparseConversion.R](https://github.com/binxiangni/RcppArmadillo/blob/master/inst/unitTests/runit.sparseConversion.R) so that the new test file can contain all tests for sparse matrix conversion?

eddelbuettel:
Thanks for switching to `text=...`.   As for moving tests, I have no strong feelings.  If they fit better by their functionality, sure.  We could also leave them by their historic.  Most important is to have tests, and we now have lots which is excellent.

binxiangni:
Some of them in the previous test file might be the same as the new ones in runit.sparseConversion.R. Maybe just put them aside now and integrate them next time.

PR #145, opened 2017-07-10
Rewrite conversion for ddiMatrix

binxiangni:
When doing tests for ddiMatrix, an extreme case came out that `drag == "N"` and `0 in x`. So I rewrite the conversion for ddiMatrix to support that.   

The unit tests for sparse matrix conversion are also moved to [runit.sparseConversion.R](https://github.com/binxiangni/RcppArmadillo/blob/master/inst/unitTests/runit.sparseConversion.R) to keep the functionality.  

When looking through the documentation of Armadillo, I find that it might not support boolean matrix (please correct me if I am wrong). The multiplication of boolean matrix is different from that of numeric matrix. So the conversion for lsparseMatrix and nsparseMatrix might not make sense and could be left to be done when Armadillo supports the operation of boolean matrices. Next I'll add warning message for lsparseMatrix and nsparseMatrix and maintain the conversion after the new feature for boolean sparse matrix comes up in Armadillo.


eddelbuettel:
Correct on lack of boolean.  We could map on the way in -- Armadillo uses `umat` for indexing.  Or we could just say "cannot do".  Unsure.

binxiangni:
I think that even if the boolean matrix can be represented in `umat`, the multiplication for `umat` might be wrong. It will regard the matrices as numeric matrices and apply multiplication of numeric matrices to them rather than that of boolean matrices. I write a small program to test it. 

```
#include <RcppArmadillo.h>

using namespace Rcpp;
using namespace arma;

// [[Rcpp::export]]
umat times(umat X, umat Y) {
  return X * Y;
}

/*** R
library(Matrix)
xtxt <- c("1 1 0 1",
          "0 1 1 0")
X <- as.matrix(read.table(text=xtxt))
dimnames(X) <- NULL
ytxt <- c("1 1 0",
                "0 1 0",
                "0 1 1",
                "1 0 0")
Y <- as.matrix(read.table(text=ytxt))
dimnames(Y) <- NULL
times(X, Y)
*/

//times(X, Y)
//     [,1] [,2] [,3]
//[1,]    2    2    0
//[2,]    0    2    1
```

As page 176th of the [note](http://www.math.fsu.edu/~pkirby/mad2104/SlideShow/s5_4.pdf) points out, the result should be 
```
     [,1] [,2] [,3]
[1,]    1    1    0
[2,]    0    1    1
```

coatless:
Perhaps contact Conrad to add a typedef for logical matrices?

https://github.com/RcppCore/RcppArmadillo/blob/2d037907658038b90e1c79292bb11767283eebec/inst/include/armadillo_bits/typedef_mat.hpp

thirdwing:
It is OK if we can't support every type. Some friendly warnings should be fine.

PR #147, opened 2017-07-12
Add a warning message for matrices that cannot be converted

binxiangni:
Although just adding the warning message can pass the tests, it will not stop the program and provide a wrong result. Should I replace `Rcpp::warning()` with `Rcpp::stop()`?

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,10 @@
+2017-07-11  Binxiang Ni  <binxiangni@gmail.com>
+
+	* inst/include/RcppArmadilloAs.h: Add a stop message for matrices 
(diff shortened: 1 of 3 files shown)


eddelbuettel:
Yes, I was thinking that too when I wrote the message over there.

@thirdwing What do you think?  `stop()` or `warn()` ?

coatless:
Should be `Rcpp::stop()`.

Prior precedent:

https://github.com/RcppCore/RcppArmadillo/blob/cc046406f34ee732a36a0887cd35f09928c6bc27/inst/include/RcppArmadilloAs.h#L613

PR #153, opened 2017-07-31
use Rcpp vignette style

eddelbuettel:
@binxiangni I updated the vignette a little, see a [(temp) copy of the pdf](http://dirk.eddelbuettel.com/tmp/RcppArmadillo-sparseMatrix.pdf) for a quick glance.

### DIFF
--- a/vignettes/RcppArmadillo-sparseMatrix.Rnw
+++ b/vignettes/RcppArmadillo-sparseMatrix.Rnw
@@ -1,15 +1,52 @@
+% !Rnw driver = highlight::HighlightWeaveLatex()
 \documentclass[10pt]{article}
 %\VignetteIndexEntry{RcppArmadillo-sparseMatrix}
+%\VignetteEngine{highlight::highlight}
 %\VignetteKeywords{R, C++, Armadillo, linear algebra, sparse matrix}
 %\VignettePackage{RcppArmadillo}
 
-\usepackage{vmargin}
-\setmargrb{1.5in}{0.75in}{1.5in}{0.75in}
+\usepackage[USletter]{vmargin}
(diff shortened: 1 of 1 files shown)


binxiangni:
Looks good. There might be an inconsistency due to me. [Subsection 2.7.3](https://github.com/RcppCore/RcppArmadillo/pull/153/commits/32996daffd5a27356d67c783470f6cd3adf490c0#diff-955b0f1df45f7bd7af41e6baabd2ad3eL405) includes `library(Matrix)`, but others don't. Maybe we can just remove it. The other part is great. You can merge it. Thank you!

eddelbuettel:
> Maybe we can just remove it.

Done!

eddelbuettel:
BTW, one downside is that you need a helper script such as the snippet below to compile the vignette.

```r
#!/usr/bin/Rscript

## use given argument(s) as target files, or else default to .Rnw files in directory
files <- if (length(commandArgs(TRUE)) == 0) dir(pattern="*.Rnw") else commandArgs(TRUE)

## convert all files from Rnw to pdf using the highlight driver
invisible(sapply(files, function(srcfile) {
    Sweave(srcfile, driver=highlight::HighlightWeaveLatex())
    tools::texi2pdf(gsub(".Rnw", ".tex", srcfile))
}))
```

binxiangni:
Not familiar with that. I just compile my original one in Rstudio. Thanks for reminding. Really learned a lot from you this summer.

eddelbuettel:
It's the non-standard (but perfectly fine and permitted) vignette we use: based on [highlight](http://www.andre-simon.de/doku/highlight/en/highlight.php).   All pretty easy on Linux, and (I presume) possible on macOS...

PR #151, opened 2017-07-29
added optional RCPP_ARMADILLO_RETURN_COLVEC_AS_VECTOR and co.

sgsokol:
I don't know how to add unit tests with RUnit so tested by hand with code snipets like the following:

#### legacy colvec
```r
> sourceCpp(code="// [[Rcpp::depends(RcppArmadillo)]]\n#include <RcppArmadillo.h>\n// [[Rcpp::export]]\narma::vec add1(arma::vec x) {return(x+1);}")
> add1(1:2)
     [,1]
[1,]    2
[2,]    3
```

#### colvec tested with COLVEC
```r
> sourceCpp(code="// [[Rcpp::depends(RcppArmadillo)]]\n#define RCPP_ARMADILLO_RETURN_COLVEC_AS_VECTOR\n#include <RcppArmadillo.h>\n// [[Rcpp::export]]\narma::vec add1v(arma::vec x) {return(x+1);}")
> add1v(1:2)
[1] 2 3
```

#### colvec tested with ANYVEC
```r
> sourceCpp(code="// [[Rcpp::depends(RcppArmadillo)]]\n#define RCPP_ARMADILLO_RETURN_ANYVEC_AS_VECTOR\n#include <RcppArmadillo.h>\n// [[Rcpp::export]]\narma::vec add1v(arma::vec x) {return(x+1);}")
> add1v(1:2)
[1] 2 3
```

#### legacy rowvec
```r
> sourceCpp(code="// [[Rcpp::depends(RcppArmadillo)]]\n#include <RcppArmadillo.h>\n// [[Rcpp::export]]\narma::rowvec add1r(arma::rowvec x) {return(x+1);}")
> add1r(1:2)
     [,1] [,2]
[1,]    2    3
```

#### rowvec tested with ROWVEC
```r
> sourceCpp(code="// [[Rcpp::depends(RcppArmadillo)]]\n#define RCPP_ARMADILLO_RETURN_ROWVEC_AS_VECTOR\n#include <RcppArmadillo.h>\n// [[Rcpp::export]]\narma::rowvec add1rv(arma::rowvec x) {return(x+1);}")
> add1rv(1:2)
[1] 2 3
```

#### rowvec tested with ANYVEC
```r
[...]

PR #154, opened 2017-08-01
vec as vector

sgsokol:
Added unit tests.
To test all combinations of macros, we need several .cpp files:
- one where no macro defined (armadillo.cpp);
- another one where RCPP_ARMADILLO_RETURN_COLVEC_AS_VECTOR and  RCPP_ARMADILLO_RETURN_ROWVEC_AS_VECTOR are defined (colrow_as_vec.cpp)
- and the last one where RCPP_ARMADILLO_RETURN_ANYVEC_AS_VECTOR is defined (any_as_vec.cpp)
That's why the function unit_test_setup() from unit.test.R had to be modified to accept character vector as first argument. This modification is transparent for previous testing codes.

### DIFF
--- a/R/unit.test.R
+++ b/R/unit.test.R
@@ -35,12 +35,14 @@ unit_test_setup <- function(file = NULL, packages = NULL) {
             }
         }
         if (!is.null(file)) {
-            if (exists("pathRcppArmadilloTests")) {
-                sourceCpp(file.path(get("pathRcppArmadilloTests"), "cpp", file))
-            } else if (file.exists(file.path("cpp", file))) {
-                sourceCpp(file.path( "cpp", file))
-            } else {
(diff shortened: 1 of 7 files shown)


eddelbuettel:
This overlaps with #151 which is not a good idea.  Either expand the existing PR, or create a new one.

I can also just paste in the addition to the header.  

Tests underway now, co-mingled with an upgrade to Armadillo 7.960.0.   Both look good.

sgsokol:
Sorry for overlapping PRs. Now I have found instructions written in tiny characters "Add more commits by pushing to the master branch on sgsokol/RcppArmadillo". Hoping it will work as expected next time.

eddelbuettel:
It happens.  Git can be complicated.  I can probably just squash this PR on top; it will be smart enough to not reapply the two already-merged commits.

PR #155, opened 2017-08-02
Comment out unused conditional statement

binxiangni:
Since the logical sparse matrix is not supported now, the conditional statement is not used. So I comment it out.

### DIFF
--- a/inst/include/RcppArmadilloWrap.h
+++ b/inst/include/RcppArmadilloWrap.h
@@ -95,14 +95,15 @@ namespace Rcpp{
         IntegerVector i(sm.row_indices, sm.row_indices + sm.n_nonzero);
         IntegerVector p(sm.col_ptrs, sm.col_ptrs + sm.n_cols+1 ) ;
 
-        std::string klass ;
-        switch( RTYPE ){
-            case REALSXP: klass = "dgCMatrix" ; break ;
-            // case INTSXP : klass = "igCMatrix" ; break ; class not exported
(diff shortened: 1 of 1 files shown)


eddelbuettel:
That seems reasonable but I am still hesitating a little because ... it also removes the 

```c++
            default:
                throw std::invalid_argument( "RTYPE not matched in conversion to sparse matrix" ) ;
```

What if we get called with an `RTYPE` other than real?  "Good enough" because we throw from the constructor when it fails for non-numeric?

binxiangni:
In the as<>():
1.  For `dgCMatrix`, `dtCMatrix`, `dsCMatrix`, `dgTMatrix`, `dtTMatrix`, `dsTMatrix`, `dgRMatrix`, `dtRMatrix` and `dsRMatrix`, they all inherit from dsparseMatrix, whose `x` is a [numeric vector](https://stat.ethz.ch/R-manual/R-devel/library/Matrix/html/dsparseMatrix-class.html). I think it should correspond to `REALSXP`;
2.  For `indMatrix` and `pMatrix`, I set `x` to be IntegerVector, which should correspond to `INTSXP`. But since the unit tests are passed since uncommented and they are wrapped to `dgCMatrix`, I think that` INTSXP` should be converted to `REALSXP`(I'm not sure for this point);
3.  For `ddiMatrix`, I set `x` to be `std::double`, which I think should be converted to `REALSXP`;
4.  For other kinds, a stop message is provided in `as<>()`.

PR #170, opened 2017-08-28
Feature/openmp off on macos

eddelbuettel:
@kevinushey : @coatless already test drove this and said it worked on his macOS box.  Could you possibly give it a spin too, maybe even on two different macOS variants (hoping / thinking you may have access to more than one).


binxiangni:
FYI, it works on my macOS.

eddelbuettel:
Thanks, @binxiangni.  That makes it 2:0.  

Are you running the same version as @coatless, assuming the newest (ElCapitan, if I have that right).

binxiangni:
10.12.6 is Sierra, the newest system.

coatless:
@eddelbuettel: the macOS version of _R_ is built on El Capitan. As I'm on Sierra, `gfortran` encourages the use of 6.3.0 instead of _R_'s 6.1.0. My override for not using `gfortran 6.1.0` was wiped as I was testing a new toolchain installer "Rtools for macOS" that installs everything together.

danfulop:
Will RcppArmadillo support OpenMP on macOS again in the future? I followed @coatless directions for making [R & OpenMP play nice](http://thecoatlessprofessor.com/programming/openmp-in-r-on-os-x/#after-3-4-0), but RcppArmadillo still has OpenMP turned off. I'm running macOS 10.12 & R 3.4.1.

A related question. Will all R packages that depend on the Rcpp suite be lacking OpenMP support on macOS for the time being, or would that only be the case sometimes. _Please excuse my lack of knowledge with regards to Rcpp and RcppArmadillo_.

coatless:
@danfulop: This is only the case for _RcppArmadillo_. Using OpenMP on macOS with _Rcpp_ is possible via `// [[Rcpp::plugins(openmp)]]`. Though, you need to be wary of the dragons that exist since _R_ is single threaded.

You can manually regain OpenMP by modifying the local install of _RcppArmadillo_'s `inline.R` to include `-fopenmp` and removing the [`#define ARMA_DONT_USE_OPENMP 1`](https://github.com/RcppCore/RcppArmadillo/blob/a97fa6f42235d98ffca7e1c5e2a168f014639688/configure.ac#L104) in [`inst/include/RcppArmadilloConfigGenerated.h`](https://github.com/RcppCore/RcppArmadillo/blob/a97fa6f42235d98ffca7e1c5e2a168f014639688/inst/include/RcppArmadilloConfigGenerated.h.in#L29)

yaccos:
Would it not be better if the `configure` script detected directly if the compiler supported openmp?

eddelbuettel:
Yes, of course it would be. In an ideal world R would do this reliably for us as it does for most other things on most platforms.  @coatless and I looked into this repeatedly, but it still difficult to impossible given the current setup particularly on macOS.  While things work pretty well on Linux, it is a sadder story on macOS.

kthohr:
@eddelbuettel - What about a check based on
```bash
apple_compiler=$($CXX --version 2>&1 | grep -i -c -e 'apple llvm')
```
Works well on my macOS system.

coatless:
Actually, @eddelbuettel the use of the `m4` routine would work in terms of allowing other compilers to work. We just haven't solved the issue of the `clang4` detection...

kthohr:
@eddelbuettel A little spaghetti code :-) See line 65 onwards:

https://github.com/kthohr/RcppArmadillo/blob/macOS-openMP/configure.ac

I then modified the existing macOS check (line 135 in that file). I've tested this using regular (Apple) clang++, and clang 6.0 and gcc 7.0 from MacPorts.

eddelbuettel:
Well I trust you @kthohr and @coatless.  If you can cook something up, and someone else validates, I'd be open.  I just don;t have a macOS box ...

danfulop:
I volunteer as an additional Mac tester

kthohr:
@danfulop - Can you try to install using:
```R
devtools:::install_github("kthohr/RcppArmadillo",ref="macOS-openMP")
```

danfulop:
it worked perfectly!

binxiangni:
It also worked on my side(macOS 10.13.4)

eddelbuettel:
@coatless and I discussed this a little.  In principle, all this is fine, and I can probably be persuaded to use the @kthohr patch. I won't be able to test it though.

And one general issue that with the 450+ reverse depends we have some responsibility to CRAN to no change this too much / require too many uploads.  But we can certainly think about folding improvements in at the so-far bi-monthly cycle.

kthohr:
I'll move this to a new PR.

Issue #161, opened 2017-08-17
Quanteda regression


eddelbuettel:
Those watching this repo and hence seeing this probably also follow the rcpp-devel list.  

The [quanteda](https://github.com/kbenoit/quanteda) by @kbenoit et al  package came up as it at times seems to fail tests for me / us / CRAN.  Now that RcppArmadillo 0.760.1.0 is on CRAN, it is seen to let quanteda fail 14 times:

```
* checking tests ... [59s/36s] ERROR
  Running ‘testthat.R’ [58s/36s]
Running the tests in ‘tests/testthat.R’ failed.
Last 13 lines of output:
  testthat results ================================================================
  OK: 1390 SKIPPED: 19 FAILED: 17
  1. Error: test textplot_scale1d wordfish in the most basic way (@test-plots.R#96) 
  2. Error: textmodel-ca works as expected as ca::ca : use mt (@test-textmodel_ca.R#27) 
  3. Error: textmodel-ca(sparse) works as expected on another dataset (@test-textmodel_ca.R#68) 
  4. Error: textmodel-wordfish works as expected: dense vs sparse vs sparse+mt (@test-textmodel_wordfish.R#22) 
  5. Error: print/show/summary method works as expected (@test-textmodel_wordfish.R#32) 
  6. Error: coef works for wordfish fitted (@test-textmodel_wordfish.R#64) 
  7. Error: textmodel-wordfish works for quasipoisson - feature as expected: dense vs sparse vs sparse+mt (@test-textmodel_wordfish.R#81) 
  8. Error: textmodel-wordfish works for quasipoisson - overall as expected: dense vs sparse vs sparse+mt (@test-textmodel_wordfish.R#103) 
  9. Error: textmodel-wordfish (sparse) works as expected on another dataset (@test-textmodel_wordfish.R#119) 
  1. ...
  
  Error: testthat unit tests failed
  Execution halted
* checking for unstated dependencies in vignettes ... OK
```

Most of these are of the `dfmSparse is not supported` variety leading to our `as<>()` caster.

For testing, I downgraded to the previously release version 0.7.900.2.0 of RcppArmadillo -- and with it quateda 0.99 (which itself only got onto CRAN after we submitted) passes.

@binxiangni Could you possibly take a quick look if either we (or quanteda) can accomodate with a simple cast?  I can take a look too but I am simply not that deep in the sparse code ...

kbenoit:
[...]

PR #171, opened 2017-08-31
Armadillo 8.100, call .sync() before accessing the internal arrays of the SpMat class

binxiangni:
I just call .sync() in as<>() and it works. But it seems a little weird that no error comes out when I don't call .sync() in wrap() before reading those internal arrays of the SpMat class.

### DIFF
--- a/inst/include/RcppArmadilloAs.h
+++ b/inst/include/RcppArmadilloAs.h
@@ -105,6 +105,9 @@ namespace traits {
                 // Making space for the elements
                 res.mem_resize(static_cast<unsigned>(x.size()));
                 
+                // In order to access the internal arrays of the SpMat class
+                res.sync();
+                
                 // Copying elements
                 std::copy(i.begin(), i.end(), arma::access::rwp(res.row_indices));
                 std::copy(p.begin(), p.end(), arma::access::rwp(res.col_ptrs));
@@ -118,6 +121,9 @@ namespace traits {
                 
                 // Making space for the elements
                 res.mem_resize(static_cast<unsigned>(x.size()));
+                
+                // In order to access the internal arrays of the SpMat class
+                res.sync();
                 
                 // Copying elements
                 std::copy(i.begin(), i.end(), arma::access::rwp(res.row_indices));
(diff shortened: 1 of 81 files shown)


eddelbuettel:
Your tests may be missing the salient fact that we are *not* on Armadill 0.8[01]00.* yet.

binxiangni:
So in addition to copying armadillo_bits to here and modifying as<>(), what else should I do?

binxiangni:
I find that I don't need to call .sync() before arma::access::rwp, but I should call that before arma::access::rw, which possibly means no need to call .sync() before directly reading or writing the pointers. I guess that might explain my problem.

eddelbuettel:
> So in addition to copying armadillo & armadillo_bits to here and modifying as<>(), what else should I do?

I think you can wait for more tests until I had a chance to update Armadillo within RcppArmadillo.

eddelbuettel:
I am not at Armadillo 8.100.1.  I think I will need the first part of the PR -- as there are some errors -- but there are "issues" with this PR as it covers 80+ files it should not cover (from Armadillo 8.*).  Shall we / can we clean that up, or shall I just re-apply the relevant part by hand?

eddelbuettel:
(Copying over.  Wrote this in #175 by mistake at first.)

This was a very important and timely PR. I had test failures due to SpMat changes, but with your syncing it passes.

Now it so happens that I cannot merge this now, but I will give you full credit in inst/NEWS.Rd. Hope that's ok.

I should have put 8.100.* into a branch earlier. My bad.

eddelbuettel:
Ok, I folded this into #176 which is now being tested.

binxiangni:
Sorry for being late to take action. Being busy with midterms. Kind of difficult to catch you up. So what should I do? As I see, you've removed the redundant files, right?

eddelbuettel:
No real action item.  But if you and @sgsokol can look over the changes I made to [`RcppArmadilloAs.h`](https://github.com/RcppCore/RcppArmadillo/blob/arma-8.100.1/inst/include/RcppArmadilloAs.h) in the branch that is behind the #161 PR.  It is now running a rev.dep -- we should be good.

binxiangni:
I checked it and my part should be fine. Thanks! (btw, the PR number above should be #176 )

eddelbuettel:
Ooops. Correcting.

sgsokol:
For me too, it looks OK.

eddelbuettel:
Again, sorry to both of for sitting on the PRs in a way that then prevented me from merging them.

Lesson learned, I just should not wait for my next RcppArmadillo release cycle.

eddelbuettel:
The content of this PR is now in master via #176.  Again my apologies for not merging this sooner (maybe into a branch with arma 8.*).

Issue #173, opened 2017-09-19
dgt2dgc is slower than native Armadillo version


sgsokol:
During my last dive in sparse matrix code, I came across multiple uses of `.push_back(...)` when building some vectors, e.g. around RcppArmadilloAs.h:178. It is used notably in the code converting dgt matrices to default dgc type (may be somewhere else too, I didn't check). Such vector building method can require big number of memory operations that are known to be relatively slow. Yet there is already native armadillo code for batch building dgc matrices from triplets ijv. My test shows that the native method is much faster on big matrices, e.g. on 3000x4000 randomly sparse matrix with about 8000 non zero entries, actual code converts such matrix in 4.5 ms while native batch construction takes only 0.9 ms on my Intel Xeon E5-2609 v2 @ 2.50GHz.
I think we could rely on native batch construction not only for the speed but also simply for avoiding code duplication.
Here is my test code:
```c++
// [[Rcpp::depends(RcppArmadillo)]]
#include <RcppArmadillo.h>
using namespace arma;

// [[Rcpp::export]]
arma::sp_mat asSpMat(SEXP S) {
    return Rcpp::as<arma::sp_mat>(S);
}
// [[Rcpp::export]]
arma::sp_mat dgt2spmatNative(Rcpp::S4 mat) {
    if (Rcpp::as<std::string>(mat.slot("class")) != "dgTMatrix")
        Rcpp::stop("matrix must be of dgTMatrix type");
    urowvec ti = mat.slot("i");
    urowvec tj = mat.slot("j");
    vec tx = mat.slot("x");
    Rcpp::IntegerVector dims = mat.slot("Dim");
    int nrow = dims[0];
    int ncol = dims[1];
    
    umat loc=join_cols(ti, tj);
    //sp_mat(add_values, locations, values, n_rows, n_cols, sort_locations = true, check_for_zeros = true)
    return sp_mat(true, loc, tx, nrow, ncol, true, false);
}

/*** R
library(Matrix)
library(Rcpp)
library(microbenchmark)
nrow=3000
ncol=4000
nnz=2*ncol
i=sample(1:nrow, nnz, replace=TRUE)
j=sample(1:ncol, nnz, replace=TRUE)
set.seed(7)
x=rnorm(nnz)
x[1]=0. # see if it's kept in conversions
dgt=sparseMatrix(i = i, j = j, x=x, dims=c(nrow, ncol), giveCsparse=FALSE)
[...]

PR #174, opened 2017-09-20
dgt, dtt and dst types converted by native batch constructor

sgsokol:
Here we go with PR for issue #173 ... but I am afraid to make again not very clean PR as it includes (I don't know why) inst/NEWS.Rd which I didn't touch. The modified version of NEWS.rd came from my earlier syncro with the main repo.
Anyhow, the patch adds very little of code (but removes much more) so if the PR is not acceptable as it is, may be copy/paste will be possible option?
On my side, unit tests pass OK and timing measurements give similar results both for asSpMat(dgt) and dgt2spmatNative(dgt) from the code posted in #173.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: RcppArmadillo
 Type: Package
 Title: 'Rcpp' Integration for the 'Armadillo' Templated Linear Algebra Library
-Version: 0.7.960.1.2
+Version: 0.7.960.1.2-1
 Date: 2017-08-29
 Author: Dirk Eddelbuettel, Romain Francois, Doug Bates and Binxiang Ni
 Maintainer: Dirk Eddelbuettel <edd@debian.org>


coatless:
Nicely done. However, is there anyway you could update your branch to avoid having all of these old commits? Hard to easily tell the difference.... 

```
git rebase master
git push --force-with-lease
```

eddelbuettel:
Thanks for the PR.  

A rebase, or simply restart which a fresh fork, would be good.

Also for version, just increment (or create) a new minor please, i.e `Version: 0.7.960.1.2.1` as a first revision to 0.7.960.1.2.

sgsokol:
I was preparing a new PR based on a clean re-fork when I saw that you already committed 0f3d81b. If I submit a new PR it will overlap, won't it?

sgsokol:
My bad, a green check sign means Travis CI build OK not a commit. So I continue with a new PR.

binxiangni:
Personally, I prefer to keep the `dgt`, `dtt` and `dst` separately under different `else if` to keep the consistency and readability, although it may lead to duplicate code. If you think now is good, I'll integrate other conversion.

sgsokol:
Sure, you keep the style of code that you prefer. I wanted just to bring a general idea.

coatless:
This PR is now in #174. It can be closed.

PR #176, opened 2017-10-02
Arma 8.100.1

eddelbuettel:


eddelbuettel:
Good news.  Of the 400 reverse depends, we skipped two by choice, had only one failure (which also seems to fail at CRAN) and 397 passes.

PR #188, opened 2017-11-13
update reference to current article

conradsnicta:
Ryan has made many contributions to Armadillo, so we should give him credit too.


eddelbuettel:
Yes, absolutely. I refer to the JOSS in some other places too.   Thanks for the PR.

PR #192, opened 2017-11-30
conversion of simple_triplet_matrix

sgsokol:
A small patch adding conversion of simple_triplet_matrix format from package slam.
Sorry for all these white spaces removing, it's my new editor Atom who's responsible.
If it's too annoying for you, I'll try to disable this feature and submit with only pertinent part of
modifications.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: RcppArmadillo
 Type: Package
 Title: 'Rcpp' Integration for the 'Armadillo' Templated Linear Algebra Library
-Version: 0.8.300.0.0
+Version: 0.8.300.0.1
 Date: 2017-11-22
 Author: Dirk Eddelbuettel, Romain Francois, Doug Bates and Binxiang Ni
 Maintainer: Dirk Eddelbuettel <edd@debian.org>
--- a/inst/include/RcppArmadilloAs.h
+++ b/inst/include/RcppArmadilloAs.h
@@ -83,7 +83,19 @@ namespace traits {
     template <typename T>
     class Exporter< arma::SpMat<T> > {
     public:
-        Exporter( SEXP x ) : mat(x){}
+        Exporter( SEXP x ) {
+            if (Rf_inherits(x, "simple_triplet_matrix")) {
+                List li = x;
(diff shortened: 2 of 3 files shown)


coatless:
I'm a bit worried about extending to meet all _custom_ objects defined. `Matrix` is one thing as it part of the recommended packages. Not so with `slam`.

Also, please modify this commit to respect spacing.

eddelbuettel:
I think we managed to use `slam`, `SparseM`, and `SciPy` as _optional_ pieces in the tests.  That way they do not bubble up into Depends.  I would much prefer to keep it that way if we could.

sgsokol:
@coatless 
- I don't think that taking into account slam format adds significant charge on the package. The code is small and compared to previous executions only one type check is added;
- you are right that Matrix is in recommended packages and not slam, but Matrix is much slower to load than slam. So if someone manages to get all he wants from a small package, why to bother him to force to drag a big one?
- white spaces are respected now.
@eddelbuettel  I didn't find any mention of slam in unit tests (a part in comments) so I've just put the test in `if (suppressMessage(require(slam))) {}` block.

eddelbuettel:
Correct.  They way you have the test conditioned we should only need a Suggests:.   And on the C++ side it just adds another type.  I see no harm in that.  

But as a general rule (to both (!!) of you):  I generally still prefer issue ticket discussion so that everybody can pipe in _before_ filing PRs.  But you two of course have built up _some_ trust by now, but I remain a very mean maintainer (even if I just merged @coatless 's PR to Rcpp....)

coatless:
@eddelbuettel in my defense, I didn't think you would appreciate an issue ticket for a six character change on a single line followed immediately by a PR ticket.

eddelbuettel:
To make it plain, we half a dozen communication channels between us. Pick one.  Any one.

Randomly submitting PRs is always the worse alternative.

sgsokol:
I realize that even with this patch we still need to load Matrix package to get it working. This invalidates my comment about time differences in loading Matrix and slam :(

eddelbuettel:
Oh well.  It'll be in 0.8.300.1.0 which I am currently rev.dep testing.  Some fixes by Conrad in 8.300.1.

PR #193, opened 2017-12-04
Feature/arma8.300.1

eddelbuettel:


eddelbuettel:
Passed reverse depensd with flying colours.  No errors, all passes, fived skips.

Issue #194, opened 2017-12-05
mandatory Matrix package


sgsokol:
Currently, it is mandatory to load package Matrix if someone get sparse matrix in parameter or turn it back to R environment. I proposed a patch (pr #192) that was supposed to circumvent an obligation to do so but Matrix is still needed. I would like to propose another one that will finally allow to avoid an obligation to load Matrix.
With this new path, a user function can take a simple_triplet_matrix type as argument and (to go in opposite direction) use a function 
```
List simple_triplet_matrix(const arma::SpMat<T> &)
```
to be able to return a sparse matrix without requiring S4 classes defined in Matrix package.

The motivation behind all of this is that a user needing only lightweight operations on sparse matrices should have an opportunity to pass by lightweight packages. Unfortunately, Matrix can require more than 1&nbsp;s to load which can be prohibitive for programs intended to run in very short time.

eddelbuettel:
Well but there are other conversion besides the one you added, and they all need Matrix as it is creates the dgCMatrix.  Plus, Matrix is almost base R so I am not sure I _really_ care about the extra installation footprint.

sgsokol:
I have made tests unloading Matrix and it passes.
Even being almost base R, Matrix is not loaded every time (fortunately!)

eddelbuettel:
Ok, lighter depends is a good goal.  Do the tests requiring Matrix then become optional?  I am still not quite sure what exactly you want to do (but _Big Thank You_ for starting with an issue ticket !).

sgsokol:
Well, they can become optional (it will require some rewriting) but for now I did not touch to them. As for tests we don't look for getting 1 s less, do we?

eddelbuettel:
No, run time is not a constraint.  Fewer depends is good, ideally at unchanged test coverage.

eddelbuettel:
Addressed in #196

PR #195, opened 2017-12-05
added simple_triplet_matrix() converter

sgsokol:
Address issue #194

### DIFF
--- a/inst/unitTests/runit.sparseConversion.R
+++ b/inst/unitTests/runit.sparseConversion.R
@@ -674,11 +674,15 @@ if (.runThisTest) {
         checkEquals(dgc, asSpMat(ind), msg="ind2dgC_9")
     } 
     if (suppressMessages(require(slam))) {
+        ## simple_triplet_matrix from package slam
         test.as.stm2dgc <- function() {
-            ## simple_triplet_matrix from package slam
             stm <- as.simple_triplet_matrix(diag(2))
             dgc <- as(as(stm, "matrix"), "dgCMatrix")
             checkEquals(dgc, asSpMat(stm), msg="stm2dgc")
+        }
+        test.as.stm2stm <- function() {
+            stm <- as.simple_triplet_matrix(diag(2))
+            checkEquals(stm, asStm(stm), msg="stm2stm")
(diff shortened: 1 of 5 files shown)


inst/unitTests/runit.sparseConversion.R
```diff
@@ -674,12 +674,19 @@ if (.runThisTest) {
...
+        ## simple_triplet_matrix from package slam
         test.as.stm2dgc <- function() {
-            ## simple_triplet_matrix from package slam
             stm <- as.simple_triplet_matrix(diag(2))
             dgc <- as(as(stm, "matrix"), "dgCMatrix")
             checkEquals(dgc, asSpMat(stm), msg="stm2dgc")
         }
+        test.as.stm2stm <- function() {
+            if ("package:Matrix" %in% search())
+                uppressMessages(unloadNamespace("Matrix"))
```
eddelbuettel: That is a type:  `suppressMessages()`.

I am not a fan of trying to unload packages. I don't think it works reliably.  What are you trying to do?

sgsokol: Just to check that even without Matrix loaded, the conversion stm -> arma -> stm works well.
If unloading is a concern, we can do without it but we have to have a faith that it works without Matrix.

eddelbuettel: That is how the unit tests work (eg on Travis):  if not explicitly loaded, it is not there. I'd remove the line with the typo and the preceding test.

sgsokol: But it is explicitely loaded in runit.sparseConversion.R:34
```
.runThisTest <- suppressMessages(require(Matrix))
```

eddelbuettel: That's the point. We _need_ Matrix for the existing conversions.

sgsokol: Right. That's why I have unloaded it just for one test and reloaded it immediately after. But as I already said, it is OK for me to do without unload/reload. I have checked on my side that it worked without Matrix.

PR #214, opened 2018-03-29
Update config check for Rlapack

kthohr:
From RcppCore/RcppArmadillo/issues/213

All Rlapack distributions should now trigger crippled_lapack guard.


eddelbuettel:
Thanks for also running configure.  And you're of course spot on -- I just had a few quality minutes in the car dropping family off at the airport and didn't even think of the changed logic with the banded matrix decomp now being the driving factor.  This should help indeed.

kthohr:
Yep. Might have to re-enable a `newR` switch if these routines are ever added to Rlapack... (not likely too happen soon, I would guess)

eddelbuettel:
Indeed.  And there is the open thread on the r-devel list.  Maybe once R 3.5.0 is out we can re-hash with Peter and see if there is a way to get the next batch in for R 3.6.0 next year...

PR #217, opened 2018-04-14
make configure more robust against limited shells (closes #216)

### LINKED ISSUE #216: Compiler detection failure on Ubuntu 16.04
Notice on https://rstudio.cloud: 

```sh
checking whether g++ version is sufficient... almost
configure: WARNING: Compiler self-identifies as being compliant with GNUC extensions but is not g++.
checking LAPACK_LIBS... system LAPACK found
```

yet

```
$ g++ -v 2>&1 | awk '/^.*g.. version/ {print $3}'
5.4.0
$ 
```

Fix forthcoming.

eddelbuettel:


configure
```diff
@@ -936,6 +938,15 @@ do
   | -silent | --silent | --silen | --sile | --sil)
     silent=yes ;;
 
+  -runstatedir | --runstatedir | --runstatedi | --runstated \
```
coatless: What parameters are you supplying when running `autoconf` to generate this portion?

eddelbuettel: None. I just say `autoconf`.

eddelbuettel: And it is 2.69 if that matters:

```sh
edd@rob:~/git/rcpparmadillo(master)$ autoconf --version
autoconf (GNU Autoconf) 2.69
Copyright (C) 2012 Free Software Foundation, Inc.
License GPLv3+/Autoconf: GNU GPL version 3 or later
<http://gnu.org/licenses/gpl.html>, <http://gnu.org/licenses/exceptions.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Written by David J. MacKenzie and Akim Demaille.
edd@rob:~/git/rcpparmadillo(master)$ 
```

Issue #224, opened 2018-05-06
An installation error: 'Expected in: flat namespace'


sentian:
I got a bunch of warnings and an error when installing the package from source (details below). I'm on Mac OS 10.13.4 and R 3.5.0. 

```
* installing *source* package ‘RcppArmadillo’ ...
** package ‘RcppArmadillo’ successfully unpacked and MD5 sums checked
checking whether the C++ compiler works... yes
checking for C++ compiler default output file name... a.out
checking for suffix of executables... 
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C++ compiler... yes
checking whether /usr/local/bin/g++-8 accepts -g... yes
checking how to run the C++ preprocessor... /usr/local/bin/g++-8 -E
checking whether we are using the GNU C++ compiler... (cached) yes
checking whether /usr/local/bin/g++-8 accepts -g... (cached) yes
checking whether g++ version is sufficient... yes, with OpenMP as version 8.1.0
checking for macOS... found
checking for macOS Apple compiler... not found
checking LAPACK_LIBS... R-supplied partial LAPACK found
configure: WARNING: Some complex-valued LAPACK functions may not be available
checking for OpenMP... found and suitable
configure: creating ./config.status
config.status: creating inst/include/RcppArmadilloConfigGenerated.h
config.status: creating src/Makevars
** libs
/usr/local/bin/g++-8 -std=gnu++11 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG  -I"/Library/Frameworks/R.framework/Versions/3.5/Resources/library/Rcpp/include" -I/usr/local/include  -I../inst/include -fopenmp -fPIC  -Wall -g -O2 -c RcppArmadillo.cpp -o RcppArmadillo.o
In file included from /Library/Frameworks/R.framework/Versions/3.5/Resources/library/Rcpp/include/RcppCommon.h:76,
                 from ../inst/include/RcppArmadilloForward.h:26,
                 from ../inst/include/RcppArmadillo.h:31,
                 from RcppArmadillo.cpp:22:
/Library/Frameworks/R.framework/Versions/3.5/Resources/library/Rcpp/include/Rcpp/sprintf.h: In function 'std::__cxx11::string Rcpp::sprintf(const char*, ...)':
[...]

PR #225, opened 2018-05-08
add warning for nonstandard macOS builds

kthohr:
Addresses #224. On a macOS system using gcc7:
```
* installing *source* package ‘RcppArmadillo’ ...
checking whether the C++ compiler works... yes
checking for C++ compiler default output file name... a.out
checking for suffix of executables...
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C++ compiler... yes
checking whether g++-mp-7 accepts -g... yes
checking how to run the C++ preprocessor... g++-mp-7 -E
checking whether we are using the GNU C++ compiler... (cached) yes
checking whether g++-mp-7 accepts -g... (cached) yes
checking whether g++ version is sufficient... yes, with OpenMP as version 7.3.0
checking for macOS... found
checking for macOS Apple compiler... not found
checking for clang compiler... not found
configure: WARNING: unsupported macOS build detected; if anything breaks, you keep the pieces.
checking LAPACK_LIBS... system LAPACK found
checking for OpenMP... found and suitable
configure: creating ./config.status
config.status: creating inst/include/RcppArmadilloConfigGenerated.h
config.status: creating src/Makevars
```
I also reverted a change to the m4 macro used on line 15 (`m4_esyscmgd_s`), was that an error? My version of autoconf complained...

### DIFF
--- a/R/SciPy2R.R
+++ b/R/SciPy2R.R
@@ -25,11 +25,16 @@
     type <- spmat$getformat()
     shape <- unlist(spmat$shape)
     data <- as.vector(reticulate::py_get_attr(spmat, "data"))
-    if (type == "csc") {
-        indices <- as.vector(reticulate::py_get_attr(spmat, "indices"))
-        indptr <- as.vector(reticulate::py_get_attr(spmat, "indptr"))
-        res <- new("dgCMatrix", i = indices, p = indptr, x = data, Dim = shape)
-    } else if (type == "coo") {
+    
+    # Since CSC Matrix from SciPy has been automatically converted to dgCMatrix, 
+    # the conversion for csc matrix is no longer needed.
+    
+    # if (type == "csc") {
+    #     indices <- as.vector(reticulate::py_get_attr(spmat, "indices"))
+    #     indptr <- as.vector(reticulate::py_get_attr(spmat, "indptr"))
+    #     res <- new("dgCMatrix", i = indices, p = indptr, x = data, Dim = shape)
+    # } 
+    if (type == "coo") {
         row <- as.vector(reticulate::py_get_attr(spmat, "row"))
         col <- as.vector(reticulate::py_get_attr(spmat, "col"))
         res <- new("dgTMatrix", i = row, j = col, x = data, Dim = shape)
--- a/configure
+++ b/configure
@@ -613,7 +613,6 @@ infodir
 docdir
 oldincludedir
 includedir
-runstatedir
 localstatedir
 sharedstatedir
 sysconfdir
@@ -685,7 +684,6 @@ datadir='${datarootdir}'
 sysconfdir='${prefix}/etc'
 sharedstatedir='${prefix}/com'
 localstatedir='${prefix}/var'
-runstatedir='${localstatedir}/run'
 includedir='${prefix}/include'
 oldincludedir='/usr/include'
 docdir='${datarootdir}/doc/${PACKAGE_TARNAME}'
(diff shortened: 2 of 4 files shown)


kthohr:
Hmm. The sparse matrix tests are throwing up an error:
```
ERROR in test.csc2dgc: Error in spmat$getformat : $ operator not defined for this S4 class
```

binxiangni:
I tested the csc, coo, and csr matrix and found out that only csc matrix is automatically converted to dgcMatrix, so it doesn't have any method from Python side. I checked the [changelog](https://github.com/cran/reticulate/blob/master/NEWS.md) of reticulate, which says 

> Added support for conversion between Matrix::dgCMatrix objects in R and Scipy CSC matrices in Python

That's why the test failed. I might fix it tomorrow.

kthohr:
Just added you to my fork, you should be able to push to the `macOS-clang-only` branch

eddelbuettel:
Looks good. Maybe make the line shorter: _if it breaks, you keep the pieces._ ?  And/or more visually apparent?  Or is WARNING good enough?

eddelbuettel:
Nice work.

BTW, I noticed that fedora tests (I think those are both machines by Prof Ripley) and one macOS test at CRAN now show errorrs -- see [here](https://cloud.r-project.org/web/checks/check_results_RcppArmadillo.html).

The macOS test seems to be what we ran into here.  Any idea why the `asSpMat(lm)` would seg.fault?

eddelbuettel:
One other issue: do we now need a versioned `Suggests:` on reticulate (and I am not sure one can -- `Imports:` do, `Suggests:` do not AFAIK) ?  Or make the code conditional on old versus new version?

binxiangni:
````
```{Rcpp}
#include <RcppArmadillo.h>

using namespace Rcpp;
using namespace arma;

// [[Rcpp::export]]
sp_mat testSparse(sp_mat X) {
  return X;
}
```
````

````
```{r}
library(Matrix)
library(RUnit)
m <- Matrix(c(0,0,2:0), 3,5, dimnames=list(LETTERS[1:3],NULL))
lm <- (m > 1)
checkException(testSparse(lm))
```

````

I tried to reproduce the error from `asSpMat(lm)`, but on my macOS, I didn't get segment fault. No idea how to fix it now.

**Comment by eddelbuettel** [comment]
Same here, so I really suspect something with his Fedora setup:

```r
R> sourceCpp("/tmp/pr225.cpp")

R> library(Matrix)

R> library(RUnit)

R> m <- Matrix(c(0,0,2:0), 3,5, dimnames=list(LETTERS[1:3],NULL))

R> lm <- (m > 1)

R> checkException(testSparse(lm))
Error in testSparse(lm) : lgCMatrix is not supported.
[1] TRUE
R> 
```

By the way, there is a nice trick if you add a segment `/*** R` followed by R code and closing `*/` the `sourceCpp()` will run the R snippet for you!  So I made your two pieces the code that follows below.  

Thanks for setting up a minimal example!

```cpp
#include <RcppArmadillo.h>

using namespace Rcpp;
using namespace arma;

// [[Rcpp::export]]
sp_mat testSparse(sp_mat X) {
  return X;
}

/*** R
library(Matrix)
library(RUnit)
m <- Matrix(c(0,0,2:0), 3,5, dimnames=list(LETTERS[1:3],NULL))
lm <- (m > 1)
checkException(testSparse(lm))
*/
```

PR #230, opened 2018-05-31
update-Rlapack-tests

kthohr:
Update Rlapack tests to trigger band matrix solvers.

Ref: conradsnicta/armadillo-code/issues/76


kthohr:
Yep. Should cover a Lapack routine that is missing in Rlapack.

PR #231, opened 2018-05-31
Fixes #229

### LINKED ISSUE #229: example_code = FALSE + pkgKitten yields error
The following sequence of commands fails :
```
library(RcppArmadillo)
RcppArmadillo.package.skeleton("foobar", example_code = FALSE)
```
with 

> Calling kitten to create basic package.
> Error in (function (name = "anRpackage", path = ".", author, maintainer, : unused argument (example_code = FALSE)
>  
> Error in value[[3L]](cond) : error while calling `kitten`

When `pkgKitten` is not available, `example_code = FALSE` works. When it is available, using the default `example_code = TRUE` works.

Session info is as follows:

> sessionInfo()
> R version 3.4.3 (2017-11-30)
> Platform: x86_64-apple-darwin17.4.0 (64-bit)
> Running under: macOS High Sierra 10.13.4
> 
> Matrix products: default
> BLAS: /System/Library/Frameworks/Accelerate.framework/Versions/A/Frameworks/vecLib.framework/Versions/A/libBLAS.dylib
> LAPACK: /System/Library/Frameworks/Accelerate.framework/Versions/A/Frameworks/vecLib.framework/Versions/A/libLAPACK.dylib
> 
> locale:
> [1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8
> 
> attached base packages:
> [1] stats     graphics  grDevices utils     datasets  methods  
> [7] base     
> 
> other attached packages:
> [1] RcppArmadillo_0.8.500.0
> 
> loaded via a namespace (and not attached):
> [1] compiler_3.4.3  tools_3.4.3     yaml_2.1.19     Rcpp_0.12.17   
> [5] pkgKitten_0.1.4

solivella:
Enable use of `example_code = FALSE` argument in `RcppArmadillo.package.skeleton()` when `pkgKitten` is installed.

### DIFF
--- a/R/RcppArmadillo.package.skeleton.R
+++ b/R/RcppArmadillo.package.skeleton.R
@@ -37,13 +37,16 @@ RcppArmadillo.package.skeleton <- function(name="anRpackage", list=character(),
     skelFunName <- ifelse(haveKitten, "kitten", "package.skeleton")
     message("\nCalling ", skelFunName, " to create basic package.")
 
+    
+    
+    
     ## first let the traditional version (or the kitten alternate) do its business
     call <- match.call()
     call[[1]] <- skelFunUsed
+    if ("example_code" %in% names(call)){
+        call[["example_code"]] <- NULL	# remove the example_code argument
+    }
     if (! haveKitten) {                 # in the package.skeleton() case
-        if ("example_code" %in% names(call)){
-            call[["example_code"]] <- NULL	# remove the example_code argument
-        }
         if (fake) {
             call[["list"]] <- "Rcpp.fake.fun"
         }
--- a/R/RcppArmadillo.package.skeleton.R
+++ b/R/RcppArmadillo.package.skeleton.R
@@ -37,9 +37,6 @@ RcppArmadillo.package.skeleton <- function(name="anRpackage", list=character(),
     skelFunName <- ifelse(haveKitten, "kitten", "package.skeleton")
     message("\nCalling ", skelFunName, " to create basic package.")
 
-    
-    
-    
     ## first let the traditional version (or the kitten alternate) do its business
     call <- match.call()
     call[[1]] <- skelFunUsed


R/RcppArmadillo.package.skeleton.R
```diff
@@ -37,13 +37,16 @@ RcppArmadillo.package.skeleton <- function(name="anRpackage", list=character(),
     skelFunName <- ifelse(haveKitten, "kitten", "package.skeleton")
     message("\nCalling ", skelFunName, " to create basic package.")
 
+    
+    
+
```
eddelbuettel: Not sure we need these but that is minor.

R/RcppArmadillo.package.skeleton.R
```diff
@@ -37,13 +37,16 @@ RcppArmadillo.package.skeleton <- function(name="anRpackage", list=character(),
...
 
+    
+    
+    
     ## first let the traditional version (or the kitten alternate) do its business
     call <- match.call()
     call[[1]] <- skelFunUsed
+    if ("example_code" %in% names(call)){
+        call[["example_code"]] <- NULL	# remove the example_code argument
+    }
```
eddelbuettel: Isn't this the same as removing `example_code` from the arguments? Ie we have two cases:
- not set, moves on
- set, removes it
So now `example_code` is effectively remove.  I don't have a problem with that (I never used it) but this would need a different deprecation process.

R/RcppArmadillo.package.skeleton.R
solivella: :) For sure. I'll delete them and amend.

R/RcppArmadillo.package.skeleton.R
solivella: It is removed from the the call object, but the original argument itself is used later on (line 107 of original file).

eddelbuettel: Indeed -- good call!

PR #270, opened 2019-09-30
Portability fixes for configure

msk:
The `test` command, as well as the `[` command, is not required to
know the `==` operator. Only a few implementations like bash and some
versions of ksh support it.

When you run `test foo == foo` on a platform that does not support the
`==` operator, the result will be false instead of true. This can lead
to unexpected behavior.

### DIFF
--- a/configure
+++ b/configure
@@ -2744,7 +2744,7 @@ fi
 $as_echo_n "checking for macOS... " >&6; }
 RSysinfoName=$("${R_HOME}/bin/Rscript" --vanilla -e 'cat(Sys.info()["sysname"])')
 
-if test x"${RSysinfoName}" == x"Darwin"; then
+if test x"${RSysinfoName}" = x"Darwin"; then
     { $as_echo "$as_me:${as_lineno-$LINENO}: result: found" >&5
 $as_echo "found" >&6; }
     { $as_echo "$as_me:${as_lineno-$LINENO}: checking for macOS Apple compiler" >&5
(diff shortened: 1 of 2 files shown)


eddelbuettel:
I have memories of having fought this before, making me unlikely to wanting to return to this. `bash` it is.  I generally change these things only when CRAN asks me too, and as it works even on their Solaris machine the status quo may change,

We also recommend filing an issue ticket prior to making PRs to not waste effort.

So let's turn this then into the issue you never files:  What type of platform would require this?

msk:
Thank you for quick feedback.

NetBSD is one of the platforms requiring this change. I am a developer of [pkgsrc](https://github.com/NetBSD/pkgsrc), the package management system for NetBSD and many other OS's, and had to add this change as a local patch when I packaged RcppArmadillo for pkgsrc.

I understand your concern about potential problems a build script change like this may cause. However, this PR is not adding anything that RcppArmadillo is not already using. RcppArmadillo's `configure` uses `=` (mostly autoconf-generated) more often then `==`.

```
❯ grep "if test " configure | grep " = " | wc -l
      33
❯ grep "if test " configure | grep " == " | wc -l
       6
```

and even `configure.ac` has one:

```
❯ grep "if test " configure.ac | grep " = "
if test "${GXX}" = yes; then
```

So, changing `==` to `=` is not only for portability but also for consistency.

kevinushey:
FWIW `s1 = s2` is mandated by POSIX (https://pubs.opengroup.org/onlinepubs/009695399/utilities/test.html), while `s1 == s2` is not. So I would agree that using `=` rather than `==` is warranted for portability (even if `bash` remains the default shell as appropriate).

As far as I know, `==` is a `bash`-ism that does exactly the same thing as `=`, so there shouldn't be any risk.

eddelbuettel:
I looked more closely (now that I am back from traveling) and the changes to `configure.ac` are pretty short and sweet and straight (and affect mostly contributed code, hi @kthohr ;-) ) so I think I'll just merge this.  (And we didn't really need a change to `configure` as it gets generated anyway but doesn't hurt...)

kthohr:
https://github.com/RcppCore/RcppArmadillo/pull/170

Some of those pre-dated my PR ;-)

eddelbuettel:
;-)  No worries.

PR #271, opened 2019-10-03
Feature/arma 10.100 rc1

eddelbuettel:


dselivanov:
Any breaking changes (as release is major)? looked briefly but haven't found any.

eddelbuettel:
I ran full reverse depends over all 659 packages *as I always do* and there weren't any.  

Haven't posted the log to the rcpp-logs repo yet but will, as usual.

Issue #280, opened 2019-12-07
Install error on macOS with clang7


pat-s:
macOS 10.14.6 

I've installed clang7 following #268. Any suggestions?

```
Error: package or namespace load failed for ‘RcppArmadillo’ in dyn.load(file, DLLpath = DLLpath, ...):
 unable to load shared object '/Users/pjs/Library/R/3.6/library/00LOCK-RcppArmadillo/00new/RcppArmadillo/libs/RcppArmadillo.so':
  dlopen(/Users/pjs/Library/R/3.6/library/00LOCK-RcppArmadillo/00new/RcppArmadillo/libs/RcppArmadillo.so, 6): Symbol not found: _wrapper2_ddot_
  Referenced from: /Users/pjs/Library/R/3.6/library/00LOCK-RcppArmadillo/00new/RcppArmadillo/libs/RcppArmadillo.so
  Expected in: flat namespace
 in /Users/pjs/Library/R/3.6/library/00LOCK-RcppArmadillo/00new/RcppArmadillo/libs/RcppArmadillo.so
Error: loading failed
```

eddelbuettel:
Which version? The "just published" CRAN version 0.9.800.3.0?  Or a different one?

`wrapper2` points to a Fortran macro [from this file](https://github.com/RcppCore/RcppArmadillo/blob/master/inst/include/armadillo_bits/compiler_setup.hpp) -- can you poke a little?

/cc @conradsnicta

conradsnicta:
@pat-s Uninstall the system level Armadillo, and then reinstall RcppArmadillo.  For more info see related issue: https://github.com/RcppCore/RcppArmadillo/issues/264

You do not need to install Armadillo in order to use RcppArmadillo.

pat-s:
Thanks. Didn't find #264 when browsing the error message.

While this worked for me, I wonder if there is a better long-term approach to this. Didn't sound like there is such a solution in #264. 

Also this only applies when people install everything from source on macOS or if the binary is not yet built for a new release, so probably not that much of a pressing thing.

Issue #264, opened 2019-08-23
Need to uninstall Armadillo to be able to build RcppArmadillo


luispuerto:
Hey! 

As you can see in [this gist][gist] is I build RcppArmadillo with Armadillo from Homebrew in my system fails to install, while if I uninstall it `brew uninstall --ignore-dependencies armadillo` and I don't have any Armadillo in my system it build normally and after I reinstall Armadillo it seems to work properly. 

I've been checking the [FAQ][] and [this post][coatless] from @coatless and I haven't understood why this is happening. 

I've also linked in `.Renviron` to `PATH="/usr/local/opt/llvm/bin:${PATH}"` 

I have install [R this way][rinstall] as @btskinner has perfectly summarize. 

Seems that RcppArmadillo can't find OpenMP since `checking for OpenMP… missing`. 

Rcpp builds without any issue. 

Any help is really welcomed. 

[gist]: https://gist.github.com/luispuerto/2d300e05b4b1de0de50cb5df84694d74
[FAQ]: https://cloud.r-project.org/web/packages/Rcpp/vignettes/Rcpp-FAQ.pdf
[coatless]: https://thecoatlessprofessor.com/programming/r-compiler-tools-for-rcpp-on-macos/ 
[rinstall]: https://www.btskinner.io/code/install-r-with-openblas-and-openmp-on-macos-mojave/

eddelbuettel:
I am not sure if  that is really an issue in RcppArmadillo -- my Debian and Ubuntu machines tend to have Armadillo also has a system library -- and not simply something oddly configured for macOS.

In any event, I cannot help you as I don't use / have access to macOS.

amirmasoudabdol:
I have similar issue on Debian 9! I think it has to do with some of Armadillo's configurations but this happens on both macOS and Debian on a cluster that I'm using. In different system, I get different wrapper issues though. 

```
Error: package or namespace load failed for 'RcppArmadillo' in dyn.load(file, DLLpath = DLLpath, ...):
 unable to load shared object '/home/name/anaconda3/lib/R/library/00LOCK-RcppArmadillo/00new/RcppArmadillo/libs/RcppArmadillo.so':
  /home/name/anaconda3/lib/R/library/00LOCK-RcppArmadillo/00new/RcppArmadillo/libs/RcppArmadillo.so: undefined symbol: wrapper_dgesv_
Error: loading failed
Execution halted
ERROR: loading failed
[...]

PR #281, opened 2019-12-09
workaround to mitigate possible interference from a system-level installation of Armadillo

conradsnicta:
related to https://github.com/RcppCore/RcppArmadillo/issues/280 and https://github.com/RcppCore/RcppArmadillo/issues/264

### DIFF
--- a/inst/include/RcppArmadilloForward.h
+++ b/inst/include/RcppArmadilloForward.h
@@ -43,6 +43,9 @@
 // DESCRIPTION file) and/or defining #define-ing ARMA_USE_CXX11_RNG
 #define ARMA_RNG_ALT         RcppArmadillo/Alt_R_RNG.h
 
+// workaround to mitigate possible interference from a system-level installation of Armadillo
+#define ARMA_DONT_USE_WRAPPER
+
 #include <armadillo>
 
 /* forward declarations */


conradsnicta:
The underlying issue stems from `#include <armadillo>`.  It seems that under clang, the current directory is not searched first (ie. the directory where `RcppArmadillo.h` lives).

Instead, clang probably first searches directories like `/usr/include/`, `/usr/local/include` and whatever the usual places are on macOS. This in turn would pick up the system-level installation of Armadillo first, instead of the Armadillo embedded within RcppArmadillo.

CC: @coatless

eddelbuettel:
Slightly confused here:  If it is only for `clang` (and possibly only for the OS from Cupertino) then why would we undef unconditionally for all use cases?

conradsnicta:
It doesn't hurt in all use cases.  There are two possible compilation paths: (1) either the system-level of Armadillo is picked up,  or (2) the Armadillo embedded with RcppAramdillo is picked up.

In both cases you do not want the wrapper to be enabled within the context of RcppArmadillo.  (The wrapper is the armadillo run-time library).

In case (2) not using the wrapper is the default operation, so `#define ARMA_DONT_USE_WRAPPER` doesn't do anything.

In case (1) we need to ensure the wrapper is not enabled, so `#define ARMA_DONT_USE_WRAPPER` is helpful.

coatless:
The solution proposed by @conradsnicta fixes the compilation issue if `armadillo` is installed system-wide via `brew`. I'd recommend merging based on @conradsnicta's https://github.com/RcppCore/RcppArmadillo/pull/281#issuecomment-563046204 explanation.

For posterity's sake, we have: 

<details>
<summary>Compiling RcppArmadillo package from source *with* fix. (Clean build)</summary>

```bash
** R
** inst
** byte-compile and prepare package for lazy loading
** help
*** installing help indices
** building package indices
** installing vignettes
** testing if installed package can be loaded from temporary location
** checking absolute paths in shared objects and dynamic libraries
** testing if installed package can be loaded from final location
** testing if installed package keeps a record of temporary installation path
* DONE (RcppArmadillo)
```
</details>

<details>
<summary>Compiling the RcppArmadillo package from source *without* proposed define fix. (Error build) </summary>

Mirrors #280 error.

```bash
Error: package or namespace load failed for ‘RcppArmadillo’ in dyn.load(file, DLLpath = DLLpath, ...):
 unable to load shared object '/Library/Frameworks/R.framework/Versions/3.6/Resources/library/00LOCK-RcppArmadillo/00new/RcppArmadillo/libs/RcppArmadillo.so':
  dlopen(/Library/Frameworks/R.framework/Versions/3.6/Resources/library/00LOCK-RcppArmadillo/00new/RcppArmadillo/libs/RcppArmadillo.so, 6): Symbol not found: _wrapper2_ddot_
  Referenced from: /Library/Frameworks/R.framework/Versions/3.6/Resources/library/00LOCK-RcppArmadillo/00new/RcppArmadillo/libs/RcppArmadillo.so
  Expected in: flat namespace
 in /Library/Frameworks/R.framework/Versions/3.6/Resources/library/00LOCK-RcppArmadillo/00new/RcppArmadillo/libs/RcppArmadillo.so
Error: loading failed
Execution halted
ERROR: loading failed
```

</details>

PR #294, opened 2020-05-05
check for systems pre-configured to use OpenMP

### LINKED ISSUE #290: RcppArmadillo with macOS and OpenMP In a Post R 4.0.0 World
From Simon:

> we are now building with macOS 10.13 (High Sierra) as target (i.e. the oldest supported system), regular Apple Xcode/command line tools and GNU Fortran 8.2. R builds are running on macOS 10.15 (Catalina) with Xcode 11.4 using macOS 10.13 target. Packages are built on macOS 10.13 VMs with just Apple command line tools (this should make it easy to replicate the setup using Travis, for example). All 3rd party libraries that CRAN uses are available in http://mac.r-project.org/libs-4/

Source: https://stat.ethz.ch/pipermail/r-sig-mac/2020-April/013289.html

This means that macOS is no longer supporting a compiler compatible with OpenMP by default. So, custom compilations of RcppArmadillo will be required to enable OpenMP. I don't think that RcppArmadillo will be affected by this change as it already disables OpenMP when `which clang` returns an Apple identifier.

https://github.com/RcppCore/RcppArmadillo/blob/284d3182ed3a42d4e77b2220dda0eda31f767578/configure.ac#L74

Though, that should be verified on an `r-devel` build of macOS.

kevinushey:
Closes #290.

With this PR, `RcppArmadillo` should enable OpenMP if it detects that OpenMP appears to work "out of the box", without any extra flags. Output on my machine:

```
kevinushey@Kevins-MBP:~/r/pkg/RcppArmadillo [feature/openmp-detect-macos]
$ R CMD config CPPFLAGS && R CMD config LDFLAGS && R CMD INSTALL --preclean .
-I/usr/local/opt/libomp/include -Xclang -fopenmp
-L/usr/local/lib -L/usr/local/opt/libomp/lib -lomp
* installing to library ‘/Users/kevinushey/Library/R/4.0/library’
* installing *source* package ‘RcppArmadillo’ ...
** using staged installation
checking whether the C++ compiler works... yes
checking for C++ compiler default output file name... a.out
checking for suffix of executables...
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C++ compiler... yes
checking whether clang++ -mmacosx-version-min=10.13 -std=gnu++11 accepts -g... yes
checking how to run the C++ preprocessor... clang++ -mmacosx-version-min=10.13 -std=gnu++11 -E
checking whether we are using the GNU C++ compiler... (cached) yes
checking whether clang++ -mmacosx-version-min=10.13 -std=gnu++11 accepts -g... (cached) yes
checking whether R can compile programs using OpenMP already... yes
checking LAPACK_LIBS... R-supplied partial LAPACK found
configure: WARNING: Some complex-valued LAPACK functions may not be available
configure: creating ./config.status
config.status: creating inst/include/RcppArmadilloConfigGenerated.h
config.status: creating src/Makevars
** libs
clang++ -mmacosx-version-min=10.13 -std=gnu++11 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG  -I'/Users/kevinushey/Library/R/4.0/library/Rcpp/include' -I/usr/local/opt/libomp/include -Xclang -fopenmp  -I../inst/include  -fPIC  -Wall -g -O2  -c RcppArmadillo.cpp -o RcppArmadillo.o
clang++ -mmacosx-version-min=10.13 -std=gnu++11 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG  -I'/Users/kevinushey/Library/R/4.0/library/Rcpp/include' -I/usr/local/opt/libomp/include -Xclang -fopenmp  -I../inst/include  -fPIC  -Wall -g -O2  -c RcppExports.cpp -o RcppExports.o
clang++ -mmacosx-version-min=10.13 -std=gnu++11 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG  -I'/Users/kevinushey/Library/R/4.0/library/Rcpp/include' -I/usr/local/opt/libomp/include -Xclang -fopenmp  -I../inst/include  -fPIC  -Wall -g -O2  -c fastLm.cpp -o fastLm.o
clang++ -mmacosx-version-min=10.13 -std=gnu++11 -dynamiclib -Wl,-headerpad_max_install_names -undefined dynamic_lookup -single_module -multiply_defined suppress -L/Library/Frameworks/R.framework/Resources/lib -L/usr/local/lib -L/usr/local/opt/libomp/lib -lomp -o RcppArmadillo.so RcppArmadillo.o RcppExports.o fastLm.o -L/Library/Frameworks/R.framework/Resources/lib -lRlapack -L/Library/Frameworks/R.framework/Resources/lib -lRblas -L/usr/local/Cellar/gcc/9.3.0_1/lib/gcc/9/gcc/x86_64-apple-darwin19/9.3.0/../../.. -F/Library/Frameworks/R.framework/.. -framework R -Wl,-framework -Wl,CoreFoundation
installing to /Users/kevinushey/Library/R/4.0/library/00LOCK-RcppArmadillo/00new/RcppArmadillo/libs
** R
** inst
** byte-compile and prepare package for lazy loading
** help
*** installing help indices
** building package indices
** installing vignettes
** testing if installed package can be loaded from temporary location
** checking absolute paths in shared objects and dynamic libraries
** testing if installed package can be loaded from final location
** testing if installed package keeps a record of temporary installation path
* DONE (RcppArmadillo)
```

This PR works by just calling `R CMD SHLIB` on a C++ script using OpenMP. If we can compile and link that file directly, then we can skip the other OpenMP checks.


.gitignore
```diff
@@ -1,6 +1,8 @@
 .Rproj.user
 .Rhistory
 .RData
+inst/include/RcppArmadilloConfigGenerated.h
```
kevinushey, line 4: Because these are generated by their corresponding `.in` files, I think we can just `.gitignore` them.

configure.ac
```diff
@@ -19,20 +19,52 @@ AC_INIT(RcppArmadillo, m4_esyscmd_s([awk -e '/^Version:/ {print $2}' DESCRIPTION
...
+## any extra hand-holding?
+openmp_already_works="no"
 
 ## default to not even thinking about OpenMP as Armadillo wants a pragma
 ## variant available if and only if C++11 is used with g++ 5.4 or newer
-can_use_openmp=""
+can_use_openmp="no"
+
+## Check if R is configured to compile programs using OpenMP out-of-the-box.
+AC_MSG_CHECKING([whether R can compile programs using OpenMP already])
```
eddelbuettel, line 42: That trailing 'already' looks a little wooden, I may reword that 'eventually',

configure.ac
```diff
@@ -19,20 +19,52 @@ AC_INIT(RcppArmadillo, m4_esyscmd_s([awk -e '/^Version:/ {print $2}' DESCRIPTION
 if test -z "${R_HOME}"; then
     AC_MSG_ERROR([Could not determine R_HOME.])
 fi
+
 ## Use R to set CXX and CXXFLAGS
+R=${R_HOME}/bin/R
```
eddelbuettel, line 24: One can do this, and then use `R` as below, I found it helpful to to keep the longer `${R_HOME}/bin/R` to remind myself 
of the many time KH had to remind me to not use use `R` from the path.  So weak personal preference for the longer form.

eddelbuettel:
It also leaves three files around, we may want to remove those straight from `configure.ac` once 
the right variable is set.

PR #295, opened 2020-05-05
Reword, indent, remove temporaries, roll minor

eddelbuettel:
Nothing major here but we now 
- show the (large) indented block for better logic, 
- add `--clean` to `R CMD SHLIB` (catching just the .o file)
- explicitly remove .cpp, so, .dylib (does that catch all cases?)
- add ChangeLog and NEWS as usual

I wonder if we could ask `autoconf` for a temp. dir.  Probably cleanest.


kevinushey:
This LGTM and I think obviates the need for my other PR.

eddelbuettel:
> This LGTM and I think obviates the need for my other PR.

I am still wondering if we should something else (other than believing `TEMPDIR`, which is e.g. empty on my box) as discussed on Slack.  And/or make the filename more unique.  To be seen...  

Also, is there another possible extension besides `.so` and `.dylib` we need to worry about?

PR #296, opened 2020-05-05
work in private build dir during configure

kevinushey:
Performs `R CMD SHLIB` in a private directory within the temporary directory (from `TMPDIR` if set; otherwise just `/tmp`) and then removes that directory.


eddelbuettel:
Please see #295 which is in flight.  If you PR, can you possible PR against its branch, not `master`?

eddelbuettel:
The build dir generalization is good.  I'll try to make time (in a day or two) to resolve the conflict due to us working on the same file at the same time, and get the change in.

eddelbuettel:
Rebased, but need to check again it's all the way it is supposed to be...

eddelbuettel:
@kevinushey Have a quick glance over morning coffee, this should now be "final" if we get real lucky.

kevinushey:
LGTM; thanks for tidying everything up!

eddelbuettel:
Thanks for getting the ball rolling with the PR!

Issue #314, opened 2020-10-23
v0.10.1.0.0 install error on macOS


mjsteinbaugh:
I'm seeing this error when attempting to install RcppArmadillo v0.10.1.0.0 on macOS:

```
> install.packages("RcppArmadillo", type = "source")
Installing package into ‘/usr/local/koopa/opt/r/4.0/site-library’
(as ‘lib’ is unspecified)
trying URL 'https://packagemanager.rstudio.com/all/352/src/contrib/RcppArmadillo_0.10.1.0.0.tar.gz'
Content type 'binary/octet-stream' length 1642964 bytes (1.6 MB)
==================================================
downloaded 1.6 MB

* installing *source* package ‘RcppArmadillo’ ...
** package ‘RcppArmadillo’ successfully unpacked and MD5 sums checked
** using staged installation
checking whether the C++ compiler works... yes
checking for C++ compiler default output file name... a.out
checking for suffix of executables...
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C++ compiler... yes
checking whether clang++ -mmacosx-version-min=10.13 -std=gnu++11 accepts -g... yes
checking how to run the C++ preprocessor... clang++ -mmacosx-version-min=10.13 -std=gnu++11 -E
checking whether we are using the GNU C++ compiler... (cached) yes
checking whether clang++ -mmacosx-version-min=10.13 -std=gnu++11 accepts -g... (cached) yes
checking whether we have a suitable tempdir... /var/folders/l1/8y8sjzmn15v49jgrqglghcfr0000gn/T
checking whether R CMD SHLIB can already compile programs using OpenMP... yes
checking LAPACK_LIBS... R-supplied partial LAPACK found
configure: WARNING: Some complex-valued LAPACK functions may not be available
configure: creating ./config.status
config.status: creating inst/include/RcppArmadilloConfigGenerated.h
config.status: creating src/Makevars
** libs
clang++ -mmacosx-version-min=10.13 -std=gnu++11 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG  -I'/usr/local/koopa/opt/r/4.0/site-library/Rcpp/include' -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX.sdk -I/usr/local/include  -I../inst/include  -fPIC  -Wall -g -O2  -c RcppArmadillo.cpp -o RcppArmadillo.o
In file included from RcppArmadillo.cpp:22:
[...]

PR #315, opened 2020-10-24
use quotes instead of angle brackets to include armadillo header

conradsnicta:
Related to bug https://github.com/RcppCore/RcppArmadillo/issues/314

change `#include <armadillo>` to `#include "armadillo"`

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+2020-10-24  Condrad Sanderson  <conradsnicta@users.noreply.github.com>
+
+	* inst/include/RcppArmadilloForward.h: Switch to quoted #include
+
 2020-10-21  Dirk Eddelbuettel  <edd@debian.org>
 
 	* DESCRIPTION (Version, Date): Roll minor version
--- a/inst/include/RcppArmadilloForward.h
+++ b/inst/include/RcppArmadilloForward.h
@@ -46,7 +46,7 @@
 // workaround to mitigate possible interference from a system-level installation of Armadillo
 #define ARMA_DONT_USE_WRAPPER
 
-#include <armadillo>
+#include "armadillo"
 
 /* forward declarations */
 namespace Rcpp {


eddelbuettel:
As I just wrote "Maybe. Maybe not."  I'd prefer it we would not throw quick PRs around before issue tickets.

This has stood as is on mutiple OSs, versions, compilers, upgrades, for a _decade_.  Let's be sure that this is in fact it.

conradsnicta:
> As I just wrote "Maybe. Maybe not." I'd prefer it we would not throw quick PRs around before issue tickets.
> 
> This has stood as is on mutiple OSs, versions, compilers, upgrades, for a _decade_. Let's be sure that this is in fact it.

It doesn't hurt changing `<...>` to `"..."`, and is likely to improve things.

Previous related issues: #281 #280 #264

eddelbuettel:
Can we please STOP doubling and repeating all arguments in both blaces?

eddelbuettel:
> It doesn't hurt changing <...> to "...", and is likely to improve things.

That's a conjecture.  It may, or it may not.

conradsnicta:
ok, let's analyse this.

Background. A user has two installations of Armadillo:
- One installation is in `/usr/include` (or `/usr/local/include`)
- Another installation is wherever R stores its installed packages; as an example, let's say it's in `/home/user/R/x86_64-redhat-linux-gnu-library/3.6/RcppArmadillo/include/`

Having two installations is perfectly fine, as there are two clear use cases: (i) development of pure C++ code, and (ii) development of R mixed with C++ code.

Within the installed RcppArmadillo, we have the following file: `RcppArmadilloForward.h`.  This file is stored as `/home/user/R/x86_64-redhat-linux-gnu-library/3.6/RcppArmadillo/include/RcppArmadilloForward.h`.

Within `RcppArmadilloForward.h` we have the following line:  `#include <armadillo>`

The GCC docs on the include syntax (https://gcc.gnu.org/onlinedocs/cpp/Include-Syntax.html) mention two variants:

> `#include <file>`
> This variant is used for system header files. It searches for a file named file in a _standard list of system directories_. (...)
> 
> `#include "file"`
> This variant is used for header files of your own program. It searches for a file named file _first in the directory containing the current file_, then in the quote directories and then the same directories used for `<file>`. (...)

As `#include <armadillo>` is used within `RcppArmadilloForward.h`, the compiler will first search in standard directories like `/usr/include` and `/usr/local/include`. This will pick the system installation of Armadillo, and _ignore_ the installation in `/home/user/R/x86_64-redhat-linux-gnu-library/3.6/RcppArmadillo/include/`. We don't want this.

If `#include "armadillo"` is used instead within `RcppArmadilloForward.h`, the compiler will first search in the directory where `RcppArmadilloForward.h` is stored. In this example it will be `/home/user/R/x86_64-redhat-linux-gnu-library/3.6/RcppArmadillo/include/`. This is what we want.

ChangeLog
```diff
@@ -1,3 +1,7 @@
+2020-10-24  Condrad Sanderson  <conradsand@gmail.com>
```
conradsnicta, line 1: Please remove this email address.  If an email address is really necessary, I'm okay for the associated minor change in code to be attributed to to you, or perhaps "RcppArmadillo community" in general.

eddelbuettel: No that's what we do. Changes get a ChangeLog entry.

eddelbuettel:
The discussion at the gcc site is actually longer that what you quoted and also states that it should follow the include path as specified by the `-I` switches.

Which for some reason nobody can understand seems to be ordered the wrong way in the case of R on macOS.  

The issue is still bizarre because I happen to either maintain (or use) a number of other package with a similar structure of shipping their own header files which might well conflict with a system installation -- RcppEigen and RcppSpdlog are two good examples even if the latter is pretty recent.  And _e.g._ could have / should have thrown up the same issue but as far as I remember has not.  

Oh well, on balance this appears to be the right fix so I am going to merge this.

PR #319, opened 2020-12-31
Fix the unit test error from converting ddi matrix to dgC matrix

binxiangni:
The coercion from ddi matrix to dgC matrix, e.g. `as(ddi_matrix, "dgCMatrix")`,  does not work as before, and its output is dtC matrix instead of dgC matrix.  So I have coerce again from dtC matrix to dgC matrix.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+2020-12-30  Binxiang Ni  <binxiangni@gmail.com>
+
+	* inst/tinytest/test_sparseConversion.R: Coerce dtCMatrix to dgCMatrix in one unit test
+
 2020-11-15  Dirk Eddelbuettel  <edd@debian.org>
 
 	* DESCRIPTION (Version, Date): RcppArmadillo 0.10.1.2.0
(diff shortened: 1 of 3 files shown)


eddelbuettel (approved):
Very attentive, thank you. I actually made the *exact same change* myself a few days ago.

I made one more a few lines down to get rid of a warning:

```r
-T2 <- rsparsematrix(40, 12, nnz = 99, giveCsparse=FALSE)
+T2 <- rsparsematrix(40, 12, nnz = 99, repr="T")
```

eddelbuettel:
I just double checked.  The warning arises when one runs the unit test file directly (so that warnings are not swept under the rug):

```sh
edd@rob:~/git/rcpparmadillo(pr/319)$ Rscript -e 'tinytest::run_test_file("inst/tinytest/test_sparseConversion.R")'
Running test_sparseConversion.R.......   86 tests OK 1.2s
All ok, 86 results (1.2s)
Warning message:
In sparseMatrix(i = ijI%%nrow, j = ijI%/%nrow, index1 = FALSE, symmetric = symmetric,  :
  'giveCsparse' has been deprecated; setting 'repr = "T"' for you
edd@rob:~/git/rcpparmadillo(pr/319)$ 
```

PR #320, opened 2020-12-31
Replace the checkException with equal_error function

binxiangni:
If I don't remember wrong, we previously used `RUnit` package to conduct unit tests. But given we migrated to `tinytest` and `checkException` function is not available in the new package, the unit tests commented out failed. After replacing the functions with `equal_error` function from `tinytest` package, those unit tests work now. 

The [failure of tests](https://www.r-project.org/nosvn/R.check/r-devel-linux-x86_64-fedora-clang/RcppArmadillo-00check.html) should be due to the unit test we fixed yesterday, which lives in the line 114 of that unit test file.

### DIFF
--- a/inst/tinytest/test_sparseConversion.R
+++ b/inst/tinytest/test_sparseConversion.R
@@ -690,44 +690,40 @@ if (suppressMessages(require(slam))) {
     }
 }
 
-## this fails persistently at CRAN on the Fedora box, but shouldn't
-## giving up and commenting out
-## https://www.r-project.org/nosvn/R.check/r-devel-linux-x86_64-fedora-clang/RcppArmadillo-00check.html
 ## test.stop <- function() {
-##     ## [Matrix] p87 (lgCMatrix)
-##     m <- Matrix(c(0,0,2:0), 3,5, dimnames=list(LETTERS[1:3],NULL))
-##     lm <- (m > 1)
-##     checkException(asSpMat(lm))
+## [Matrix] p87 (lgCMatrix)
+m <- Matrix(c(0,0,2:0), 3,5, dimnames=list(LETTERS[1:3],NULL))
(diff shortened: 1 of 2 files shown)


inst/tinytest/test_sparseConversion.R
```diff
@@ -690,44 +690,40 @@ if (suppressMessages(require(slam))) {
...
-## https://www.r-project.org/nosvn/R.check/r-devel-linux-x86_64-fedora-clang/RcppArmadillo-00check.html
 ## test.stop <- function() {
-##     ## [Matrix] p87 (lgCMatrix)
-##     m <- Matrix(c(0,0,2:0), 3,5, dimnames=list(LETTERS[1:3],NULL))
-##     lm <- (m > 1)
-##     checkException(asSpMat(lm))
+## [Matrix] p87 (lgCMatrix)
+m <- Matrix(c(0,0,2:0), 3,5, dimnames=list(LETTERS[1:3],NULL))
+lm <- (m > 1)
+expect_error(asSpMat(lm))
```
eddelbuettel, line 697: Careful. If memory serves I went through several rounds of uploads and test "tweaking" yet it still failed at the sometimes very idiosyncratic setup of these boxen by BDR.

I am not in favour of turning these tests back on.  AFAICT this is not an issue about testing exceptions or errors between `RUnit` and `tinytest`.

binxiangni: I see. Then let me close this PR.

eddelbuettel:
In such cases, a bit of discussion in an issue ticket before sending a PRs can help to set the stage and clarify things.

binxiangni:
> In such cases, a bit of discussion in an issue ticket before sending a PRs can help to set the stage and clarify things.

I will do that next time.

PR #321, opened 2021-01-06
fix sign() to handle NaN

conradsnicta:
This commit fixes the sign() function to properly handle NaN values.  
When NaN is given, it now correctly returns NaN instead of incorrectly returning zero.  
Related bug report: https://gitlab.com/conradsnicta/armadillo-code/-/issues/169

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+2021-01-06  Conrad Sanderson  <conrad.sanderson.nospam@nogmail.com>
+
+        * inst/include/armadillo_bits/arma_cmath.hpp: fix handling of NaN by arma::sign()
+
 2021-01-01  Dirk Eddelbuettel  <edd@debian.org>
 
         * .github/workflows/ci.yaml: Add CI runner using r-ci
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,4 +1,4 @@
-2021-01-06  Conrad Sanderson  <conrad.sanderson.nospam@nogmail.com>
+2021-01-06  Conrad Sanderson  <cs.rcpparmadillo@gmail.com>
 
         * inst/include/armadillo_bits/arma_cmath.hpp: fix handling of NaN by arma::sign()
 
--- a/inst/include/armadillo_bits/arma_cmath.hpp
+++ b/inst/include/armadillo_bits/arma_cmath.hpp
@@ -183,7 +183,7 @@ constexpr
 typename arma_real_only<eT>::result
 arma_sign(const eT x)
   {
-  return (x > eT(0)) ? eT(+1) : ( (x < eT(0)) ? eT(-1) : eT(0) );
+  return (x > eT(0)) ? eT(+1) : ( (x < eT(0)) ? eT(-1) : ((x == eT(0)) ? eT(0) : x) );
   }
 
 


eddelbuettel:
I am sorry but why do you suggest all of a sudden to path your code in our package?  I mean :man_shrugging: sure why not but why not rely on what we have for ten years and have your release signify your release?

eddelbuettel:
And of course, don't get me wrong a bug is a bug and a bug needs fixing but ... around here people look first for the Rcpp (Sugar) vectorised functions, and that one works:

```r
> library(Rcpp)
> cppFunction("arma::vec arma_sign(arma::vec d) { return arma::sign(d); }", depends="RcppArmadillo")
> sapply(c(-1,0,1,NA,NaN), arma_sign)
[1] -1  0  1  0  0
> 
> cppFunction("NumericVector rcpp_sign(NumericVector d) { NumericVector x(sign(d)); return x;}")
> sapply(c(-1,0,1,NA,NaN), rcpp_sign)
[1] -1  0  1 NA NA
>  
```

So I think we can just wait this out to the next upgrade.

conradsnicta:
1. The `rcpp_sign()` function is not `arma::sign()`.  The `rcpp_sign()` function doesn't work with armadillo vectors and matrices.
2. As discussed via email late last year, we agreed to reduce the churn in RcppArmadillo releases (ie. keep RcppArmadillo at 0.10.1.x for a long while).  I stated that I will send you bug fixes directly as pull requests via GitHub. This is one of the fixes.

eddelbuettel:
Dear Conrad, please consider:

Re 1:  Yes, and R users almost always call `Rcpp::` functions first.  Which is why this PR, while technically helpful, may not matter.

Re 2:  We have posted some markdown files in various places and stated our position repeatedly: a *strong* preference for issue tickets and discussion leading some consensus before unsolicited PRs.  Now, granted, this PR is fairly local and focussed but .... this is simply not the right way to go about things.  

Especially as every other library I cover as port or integration to R via Rcpp works via their own well managed git repsository.  Yours was always a little, ahem, "special" with your insistence on tarballs with overwritten file timestamps etc.  Your repo, your rules, so and some point years ago I gave up pointing this out.  This here, however, is our repo so I would appreciate if you could play by its rules.  And issue ticket pointed to a commit of yours (over at GitLab) may suffice and make life a lot easier for you too.

eddelbuettel:
Also as discussed last time, it is preferred that PRs come with a corresponding ChangeLog entry in the standard format identifying the author, file, and purpose.

conradsnicta:
> And issue ticket pointed to a commit of yours (over at GitLab) may suffice and make life a lot easier for you too.

Below is a quote from the linked bug report, https://gitlab.com/conradsnicta/armadillo-code/-/issues/169.  The link was given in the pull request at the top.

> reported by Pietro Peronio  
> Sign function returns 0 when processing a NaN  
>  
>  The sign function included in Armadillo C++ returns 0 when a NaN is given as argument. I would expect to return NaN instead, like Matlab, Python's Numpy library and other softwares do.
>    #include <armadillo>
>    
>    `int main() {`  
>    `    std::cout << arma::sign(3) << std::endl; // Returns +1`   
>    `    std::cout << arma::sign(-3) << std::endl; // Returns -1`  
>    `    std::cout << arma::sign(0) << std::endl; // Returns 0`  
>    `    std::cout << arma::sign(arma::datum::nan) << std::endl; // Returns 0 instead of NaN`  
>    `    return 0;`  
>    `}`  

My own reply to the reported issue:

> fixed in: [b21cff93](https://gitlab.com/conradsnicta/armadillo-code/-/commit/b21cff932d5080bcea95c970e67e0015b3722146)

The contents of that fix is exactly the same as this PR.  In other words,
the line

    return (x > eT(0)) ? eT(+1) : ( (x < eT(0)) ? eT(-1) : eT(0) );

is changed to:

    return (x > eT(0)) ? eT(+1) : ( (x < eT(0)) ? eT(-1) : ((x == eT(0)) ? eT(0) : x) );

conradsnicta:
> Also as discussed last time, it is preferred that PRs come with a corresponding ChangeLog entry in the standard format identifying the author, file, and purpose.

Done.

ChangeLog
```diff
@@ -1,3 +1,7 @@
+2021-01-06  Conrad Sanderson  <conrad.sanderson.nospam@nogmail.com>
```
eddelbuettel, line 1: Appreciate the effort but not a valid email address.

eddelbuettel (changes requested):

ChangeLog
conradsnicta: fixed

conradsnicta:
Should be good to go now.

ChangeLog
```diff
@@ -1,4 +1,4 @@
-2021-01-06  Conrad Sanderson  <conrad.sanderson.nospam@nogmail.com>
+2021-01-06  Conrad Sanderson  <cs.rcpparmadillo@gmail.com>
```
eddelbuettel, line 1: The point is to have an address that identifies you in case someone wants to get hold of you.  So, no, a freshly created tar pit will not do.  

You are not required to file PRs here.  If you don;t want to fine.  But please respect that I have adhered to very common (if GNU-ish) standards here for well over a decade, and I am not going to stop now.

conradsnicta: It's not a "tar pit".  That's a perfectly fine email address which is valid and works. Emails sent to that address will get to me.  

The point of this email address is to have a separate and unique channel for RcppArmadillo related stuff.

eddelbuettel: Well, I am repo maintainer and I take the other view here.

eddelbuettel:
Incorporporated this by hand in https://github.com/RcppCore/RcppArmadillo/commit/db84fc2c2adbb9ad2fa764a4fe898b504fef50ff so PR no redundant --> closing

PR #332, opened 2021-04-12
Armadillo 10.4.0 and RcppArmadillo 0.10.4.0.0

eddelbuettel:


inst/include/armadillo_bits/arma_rng.hpp
```diff
@@ -480,40 +480,46 @@ struct arma_rng::randn
...
     #if defined(ARMA_RNG_ALT)
       {
-      for(uword i=0; i < N; ++i)  { mem[i] = eT( arma_rng_alt::randn_val() ); }
+      // NOTE: old method to avoid regressions in user code that assumes specific sequence
+
+      uword i, j;
+
+      for(i=0, j=1; j < N; i+=2, j+=2)  { arma_rng_alt::randn_dual_val( mem[i], mem[j] ); }
+
+      if(i < N)  { mem[i] = eT( arma_rng_alt::randn_val() ); }
```
coatless: Trying to grok why the change was needed here. (Apologies if I missed another area, the commit had a lot of space-noise.) 

The new loop seems to be hitting every second element in `arma_rng_alt::randn_dual_val`?
Also, the `if()` statement is out of the loop but is set to trigger only if `N` is an odd number?

eddelbuettel: Use an earlier RcppArmadillo up to 0.10.3.0.0 in the drar repo and run `set.seed(123); arma::randn(2,2)`.  The matrix under the _initial_ 0.10.4.0.0 was different.   So packages `colorednoise` and `GAS` failed.  

With the change the matrix is as was before, and the two packages pass like everything else.

eddelbuettel: Also, as you know, Conrad doesn't 'hang out' here so if the question was for him maybe ask [at this Gitlab commit](https://gitlab.com/conradsnicta/armadillo-code/-/commit/a6b42aad6a2eab50d2f1fac4bfea895fdf9891d0) instead.

eddelbuettel: It appears to be the exact code from an older routine `fill_simple` and yes, the `if` appears to catch the 'extra' element if an uneven number was requested.

Issue #351, opened 2021-11-02
Loss of dim 2 when using field as input


BerriJ:
HI,

consider these 2 functions:

```
#include <RcppArmadillo.h>
using namespace arma;

// [[Rcpp::export]]
field<mat> fieldtest1()
{
    field<mat> out(5, 6);

    return out;
}

// [[Rcpp::export]]
field<mat> fieldtest2(field<mat> x)
{
    return x;
}
```

Now consider this R code:

```
[R] field <- fieldtest1() 
    dim(field) 
[1] 5 6
[R] field2 <- fieldtest2(field) 
    dim(field2)                                                       
[1] 30  1
```

`fieldtest1` returns a nested list with dimensions 5,6 as expected. But `fieldtest2` just imports and exports the field drop dimension 2 and returns a 30,1 list. 

This Issue may be related to https://github.com/RcppCore/RcppArmadillo/issues/263.

eddelbuettel:
Thanks -- and I also think there is little new here that we didn't already have in #263.   It would appear few people work with `fields` as a data structure so it isn't biting too many people.  If it bites you and you have a moment or two I'd be happy to guide you through trying to write a PR that may fix things.  (Usual caveat that we then still have to see if it upsets existing packages, but in this case the chances may be good...)

BerriJ:
I'm happy to write a PR for this. Do you have an idea how to fix this or can you direct me to the relevant files of RcpArmadillo?

eddelbuettel:
Yes. This should in the files in `inst/include/Rcpp*` which cover the `as<>` and `wrap()` converters.  There is a `FieldImporter` in `RcppArmadilloWrap.h`.  And `RcppArmadilloAs.h` also has a bit.  Maybe you can start and poke and print some diagnostics and what came in and how.  It's been on my TODO list for years but as I have zero code using `arma::field` ...

eddelbuettel:
BTW there is a branch new branch for Armadillo 10.7.2 which is currently being bulk-tested on the 900+ reverse depends.  May as well fork off that, or wait, or maybe just rebase again later...

BerriJ:
Currently, I do only see master [here](https://github.com/RcppCore/RcppArmadillo/branches). However, rebasing will be no problem before I send a PR.

eddelbuettel:
[...]

Issue #263, opened 2019-07-30
Field of matrices being returned with class matrix


coatless:
Stumbled across an interesting hiccup in the exporter for `arma::field<T>` (the generic vector). In particular, the attributes associated with an `arma::field<arma::mat>` shows as `matrix` under a `class(x)` call instead of `list`. The only way to obtain the `list` type is to use a `typeof()` call. 

```cpp
#include<RcppArmadillo.h>
// [[Rcpp::depends(RcppArmadillo)]]

// [[Rcpp::export]]
Rcpp::List test_case() {
  
  // Random dimensions
  int N = 5,
      P = 8,
      J = 10,
      K = 3,
      S = 2;

  // Create a 2d matrix
  arma::mat standard_2d_matrix(P, N);

  // Create a 3rd array
  arma::cube standard_3d_array(J, S, N);
  
  // Create a 4D Array
  arma::field<arma::cube> nonstandard_4d_array(N);
  
  // Fill it with cubes
  nonstandard_4d_array.fill(arma::ones<arma::cube>(J, K, S));
  
  return Rcpp::List::create(
    Rcpp::Named("d2", standard_2d_matrix),
    Rcpp::Named("d3", standard_3d_array),
    Rcpp::Named("d4", nonstandard_4d_array)
  );

}
```

```r
exporter_check = test_case() 

sapply(exporter_check, FUN = class)
#       d2       d3       d4 
# "matrix"  "array" "matrix" 
sapply(exporter_check, FUN = typeof)
#      d2       d3       d4 
# "double" "double"   "list" 
```

And the object structure:

```r
str(exporter_check)
# List of 3
# $ d2: num [1:8, 1:5] 1.49e-154 1.49e-154 2.57e-322 6.95e-310 6.95e-310 ...
# $ d3: num [1:10, 1:2, 1:5] 1.49e-154 1.49e-154 9.96e-320 1.63e-322 1.43e-322 ...
# $ d4:List of 5
#  ..$ : num [1:10, 1:3, 1:2] 1 1 1 1 1 1 1 1 1 1 ...
#  ..$ : num [1:10, 1:3, 1:2] 1 1 1 1 1 1 1 1 1 1 ...
#  ..$ : num [1:10, 1:3, 1:2] 1 1 1 1 1 1 1 1 1 1 ...
#  ..$ : num [1:10, 1:3, 1:2] 1 1 1 1 1 1 1 1 1 1 ...
[...]

PR #352, opened 2021-11-02
Fix dim problem with arma::field inputs

BerriJ:
Attempts to fix https://github.com/RcppCore/RcppArmadillo/issues/351

This correctly passes the dims of nested lists when arma::field is specified as input (so field from R to Cpp).

This implementation works with 1d and 2d fields. The generalization to 3d seems very straightforward. However, just adding a third loop and extending the dims results in an error at runtime:  "product [x] do not match the length of object [x]".

This problem with 3d arrays most probably also affects "FieldImporter" (so field from C++ to R) since the following function fails with the same error:

```
#include <RcppArmadillo.h>
using namespace arma;

// [[Rcpp::export]]
field<mat> fieldtest1()
{
    field<mat> out(5, 5, 5);

    return out;
}
```

### DIFF
--- a/inst/include/RcppArmadilloAs.h
+++ b/inst/include/RcppArmadilloAs.h
@@ -22,8 +22,6 @@
 #ifndef RcppArmadillo__RcppArmadilloAs__h
 #define RcppArmadillo__RcppArmadilloAs__h
 
-#define RCPP_ARMADILLO_FIX_FieldExporter
-
 namespace Rcpp{
 
 namespace traits {
--- a/inst/include/RcppArmadilloAs.h
+++ b/inst/include/RcppArmadilloAs.h
@@ -22,6 +22,8 @@
 #ifndef RcppArmadillo__RcppArmadilloAs__h
 #define RcppArmadillo__RcppArmadilloAs__h
 
+#define RCPP_ARMADILLO_FIX_FieldExporter
+
 namespace Rcpp{
 
 namespace traits {
@@ -30,23 +32,24 @@ namespace traits {
     class Exporter< arma::field<T> > {
     public:
         Exporter(SEXP x) : data(x){}
-
         inline arma::field<T> get() {
             size_t n = data.size();
             arma::field<T> out(n);
             # if defined(RCPP_ARMADILLO_FIX_FieldExporter)
-                arma::ivec dims = data.attr("dim");
-                if (dims.n_elem == 1)
-                {
-                    out.set_size(n);
-                }
-                else if (dims.n_elem == 2)
-                {
-                    out.set_size(dims(0), dims(1));
-                }
-                else if (dims.n_elem == 3)
-                {
-                    out.set_size(dims(0), dims(1), dims(2));
+                if(!Rf_isNull(data.attr("dim"))){
+                    arma::ivec dims = data.attr("dim");
+                    if (dims.n_elem == 1)
+                    {
+                        out.set_size(n);
+                    }
+                    else if (dims.n_elem == 2)
+                    {
+                        out.set_size(dims(0), dims(1));
+                    }
+                    else if (dims.n_elem == 3)
+                    {
+                        out.set_size(dims(0), dims(1), dims(2));
+                    }
                 }
             # endif
             for (size_t i = 0; i < n; i++)
--- a/inst/include/RcppArmadilloAs.h
+++ b/inst/include/RcppArmadilloAs.h
@@ -33,22 +33,22 @@ namespace traits {
 
         inline arma::field<T> get() {
             size_t n = data.size();
-            arma::ivec dims = data.attr("dim");
-            arma::field<T> out;
-
-            if (dims.n_elem == 1)
-            {
-                out.set_size(n);
-            }
-            else if (dims.n_elem == 2)
-            {
-                out.set_size(dims(0), dims(1));
-            }
-            else if (dims.n_elem == 3)
-            {
-                out.set_size(dims(0), dims(1), dims(2));
-            }
-
+            arma::field<T> out(n);
+            # if defined(RCPP_ARMADILLO_FIX_FieldExporter)
+                arma::ivec dims = data.attr("dim");
+                if (dims.n_elem == 1)
+                {
+                    out.set_size(n);
+                }
+                else if (dims.n_elem == 2)
+                {
+                    out.set_size(dims(0), dims(1));
+                }
+                else if (dims.n_elem == 3)
+                {
+                    out.set_size(dims(0), dims(1), dims(2));
+                }
+            # endif
             for (size_t i = 0; i < n; i++)
             {
                 out(i) = as<T>(data[i]);
--- a/inst/include/RcppArmadilloAs.h
+++ b/inst/include/RcppArmadilloAs.h
@@ -32,10 +32,26 @@ namespace traits {
         Exporter(SEXP x) : data(x){}
 
         inline arma::field<T> get() {
-            size_t n = data.size() ;
-            arma::field<T> out( n ) ;
-            for(size_t i=0; i<n; i++){
-                out[i] = as<T>(data[i]) ;
+            size_t n = data.size();
+            arma::ivec dims = data.attr("dim");
+            arma::field<T> out;
+
+            if (dims.n_elem == 1)
+            {
+                out.set_size(n);
+            }
+            else if (dims.n_elem == 2)
+            {
+                out.set_size(dims(0), dims(1));
+            }
+            else if (dims.n_elem == 3)
+            {
+                out.set_size(dims(0), dims(1), dims(2));
+            }
+
+            for (size_t i = 0; i < n; i++)
+            {
+                out(i) = as<T>(data[i]);
             }
             return out ;
         }
--- a/inst/include/RcppArmadilloWrap.h
+++ b/inst/include/RcppArmadilloWrap.h
@@ -149,7 +149,11 @@ namespace Rcpp{
     template <typename T>
     SEXP wrap( const arma::field<T>& data){
         RObject x = wrap( RcppArmadillo::FieldImporter<T>( data ) ) ;
-        x.attr("dim" ) = Dimension( data.n_rows, data.n_cols, data.n_slices ) ;
+        # if defined(RCPP_ARMADILLO_FIX_FieldImporter)
+            x.attr("dim" ) = Dimension( data.n_rows, data.n_cols, data.n_slices ) ;
+        #else 
+            x.attr("dim" ) = Dimension( data.n_rows, data.n_cols ) ;
+        #endif
         return x ;
     }
 
--- a/inst/include/RcppArmadilloWrap.h
+++ b/inst/include/RcppArmadilloWrap.h
@@ -149,7 +149,7 @@ namespace Rcpp{
     template <typename T>
     SEXP wrap( const arma::field<T>& data){
         RObject x = wrap( RcppArmadillo::FieldImporter<T>( data ) ) ;
-        x.attr("dim" ) = Dimension( data.n_rows, data.n_cols ) ;
+        x.attr("dim" ) = Dimension( data.n_rows, data.n_cols, data.n_slices ) ;
         return x ;
     }
 
--- a/src/RcppExports.cpp
+++ b/src/RcppExports.cpp
@@ -6,6 +6,11 @@
 
 using namespace Rcpp;
 
+#ifdef RCPP_USE_GLOBAL_ROSTREAM
+Rcpp::Rostream<true>&  Rcpp::Rcout = Rcpp::Rcpp_cout_get();
+Rcpp::Rostream<false>& Rcpp::Rcerr = Rcpp::Rcpp_cerr_get();
+#endif
+
 // armadillo_version
 Rcpp::IntegerVector armadillo_version(bool single);
 RcppExport SEXP _RcppArmadillo_armadillo_version(SEXP singleSEXP) {
--- a/inst/include/RcppArmadilloConfig.h
+++ b/inst/include/RcppArmadilloConfig.h
@@ -135,5 +135,14 @@
 //#define RCPP_ARMADILLO_RETURN_ROWVEC_AS_VECTOR
 //#define RCPP_ARMADILLO_RETURN_ANYVEC_AS_VECTOR
 
+// To preserve all dims of arma::field when passing to R the following macro 
+// can be defined before including RcppArmadillo.h. 
+// see https://github.com/RcppCore/RcppArmadillo/pull/352
+// #define RCPP_ARMADILLO_FIX_FieldImporter
+
+// To preserve all dims of and arma::field input argument when passing to C++
+// the following macro can be defined before including RcppArmadillo.h.
+// see https://github.com/RcppCore/RcppArmadillo/pull/352
+// #define RCPP_ARMADILLO_FIX_FieldExporter
 
 #endif
--- a/inst/tinytest/test_rcpparmadillo.R
+++ b/inst/tinytest/test_rcpparmadillo.R
@@ -37,8 +37,9 @@ expect_equal( res[[2]][[2]], matrix(0, ncol = 5, nrow=1))#, msg = "zeros<fmat>(5
 expect_equal( res[[3]][[1]], matrix(0, ncol = 1, nrow=5))#, msg = "zeros<mat>(1,5)" )
 expect_equal( res[[3]][[2]], matrix(0, ncol = 1, nrow=5))#, msg = "zeros<mat>(1,5)" )
 
-expect_equal( res[[4]][[1]], matrix(0:3, ncol = 2, nrow=2))#, msg = "field<int>" )
-expect_equal( res[[4]][[2]], matrix(letters[1:4], ncol = 2, nrow=2))#, msg = "field<std::string>" )
+# TODO: Revisit field tests later and fix see https://github.com/RcppCore/RcppArmadillo/pull/352
+# expect_equal( res[[4]][[1]], matrix(0:3, ncol = 2, nrow=2))#, msg = "field<int>" )
+# expect_equal( res[[4]][[2]], matrix(letters[1:4], ncol = 2, nrow=2))#, msg = "field<std::string>" )
 
 
 # test.wrap.Glue <- function(){
--- a/inst/tinytest/test_rcpparmadillo.R
+++ b/inst/tinytest/test_rcpparmadillo.R
@@ -38,8 +38,8 @@ expect_equal( res[[3]][[1]], matrix(0, ncol = 1, nrow=5))#, msg = "zeros<mat>(1,
 expect_equal( res[[3]][[2]], matrix(0, ncol = 1, nrow=5))#, msg = "zeros<mat>(1,5)" )
 
 # TODO: Revisit field tests later and fix see https://github.com/RcppCore/RcppArmadillo/pull/352
-# expect_equal( res[[4]][[1]], matrix(0:3, ncol = 2, nrow=2))#, msg = "field<int>" )
-# expect_equal( res[[4]][[2]], matrix(letters[1:4], ncol = 2, nrow=2))#, msg = "field<std::string>" )
+expect_equal( res[[4]][[1]], matrix(0:3, ncol = 2, nrow=2))#, msg = "field<int>" )
+expect_equal( res[[4]][[2]], matrix(letters[1:4], ncol = 2, nrow=2))#, msg = "field<std::string>" )
 
 
 # test.wrap.Glue <- function(){


inst/include/RcppArmadilloAs.h
```diff
@@ -32,10 +32,17 @@ namespace traits {
         Exporter(SEXP x) : data(x){}
 
         inline arma::field<T> get() {
-            size_t n = data.size() ;
```
BerriJ, line 35: This does only extracts length. We want to access `.attr("dim"); `here to retain the object dimensions.

src/RcppExports.cpp
```diff
@@ -6,6 +6,11 @@
 
 using namespace Rcpp;
 
+#ifdef RCPP_USE_GLOBAL_ROSTREAM
```
BerriJ, line 9: Not shure why `devtools::build()` added these lines.

eddelbuettel: New Rcpp does that, i.e. it is in Rcpp 1.0.7 but wasn't in Rcpp last time the file was updated (which could have been a while).

inst/include/RcppArmadilloAs.h
```diff
@@ -32,10 +32,17 @@ namespace traits {
...
-            arma::field<T> out( n ) ;
-            for(size_t i=0; i<n; i++){
-                out[i] = as<T>(data[i]) ;
+            arma::ivec dims = data.attr("dim");
+            dims.resize(3);
```
eddelbuettel, line 36: Just thought, based on re-reading http://arma.sourceforge.net/docs.html#field : we don't actually know that `dims` has length three.   We could have a `field` of `vec`, or `mat`, or `cube`.

conradsnicta: The dimensions of the field are separate from the dimensions of the objects the field contains.  Each object with the field (eg. vector, matrix, cube, ...) can have its own unique size.

Say we have a 3D field of matrices: `field<mat> F(3,4,5)`. Within F, the matrix at F(0,0,0) could have a size of (2,3), while the matrix at F(1,2,3) could have a a size of (100,200).

conradsnicta: Of course, like @eddelbuettel said, the field itself could be 1D, 2D or 3D.

BerriJ:
So heres an Overview of the progress so far. Consider this Cpp Function:

```
// [[Rcpp::export]]
arma::field<arma::mat> fieldtest2(arma::field<arma::mat> x)
{
    return x;
}
```

Creating a simple 1d list of matrices works as expected:

```
[R]  A <- matrix(1, ncol = 2, nrow = 3)
[R]  B <- matrix(4, ncol = 5, nrow = 6)
[R]  L <- list(A,A,A,A,A, B,B,B)
[R]  dim(L) <- c(8,1)

[R] fieldtest2(L)                          
     [,1]      
[1,] numeric,6 
[2,] numeric,6 
[3,] numeric,6 
[4,] numeric,6 
[5,] numeric,6 
[6,] numeric,30
[7,] numeric,30
[8,] numeric,30
```

You may wonder why I set `[R]  dim(L) <- c(8,1)` and not just `[R]  dim(L) <- c(8)` but `c(8,1)` is how fields of dimension 1 get passed to R from Cpp. 

It also works for 2d lists:

```
[R] dim(L) <- c(4,2)                       

[R] fieldtest2(L)                          
     [,1]      [,2]      
[1,] numeric,6 numeric,6 
[2,] numeric,6 numeric,30
[3,] numeric,6 numeric,30
[4,] numeric,6 numeric,30
```

The question remains about how to deal with the 3D case. If we do not find a suitable solution we could fall back to the current implementation of flattening out the field (and maybe return a warning about that?).

eddelbuettel:
NB: I have NOT YET looked at your most recent change.  Consider this semi-related

Played a bit more with creating `field` objects based on your first example:

```c++
// [[Rcpp::export]]
Rcpp::List fieldtest1() {
    field<cube> zap(1,2,1);
    zap.print("zap");

    field<mat> out(2, 2);
    out.print("out");

    field<vec> boo(2);
    boo.print("boo");

    return Rcpp::List::create(Rcpp::Named("fc") = zap,
                              Rcpp::Named("fm") = out,
                              Rcpp::Named("fv") = boo);
}
```

The above works, but using, say, ` field<cube> zap(2,2,2);` fails _and I now suspect this is due because we force everything to be a matrix_ (which is actually a bigger RcppArmadillo decision I now regret, we never do the `drop=TRUE` feature of one-row or one-col returns becoming a vector).  So maybe by relaxing this for fields and just returning arrays we would be in business?

There are three tentative `#define` statements at the end of `RcppArmadilloConfig.h`

```c++
// To return arma::vec or arma::rowvec as R vector (i.e. dimensionless),
// one of the following macro can be defined before including
// RcppArmadillo.h. "ANYVEC" applys for both col- and row-vec.
//#define RCPP_ARMADILLO_RETURN_COLVEC_AS_VECTOR
//#define RCPP_ARMADILLO_RETURN_ROWVEC_AS_VECTOR
//#define RCPP_ARMADILLO_RETURN_ANYVEC_AS_VECTOR
```

We could see if altering their value helps.

BerriJ:
Thanks again for your comments. I just pushed some additional changes and now it works.

Previously I missed out on RcppArmadilloWrap.h but[ adding `n_slices` there](https://github.com/RcppCore/RcppArmadillo/pull/352/commits/d64d20522c733acad1afdcfb1b32d9905d59e5dc#diff-996157f07edb794a5b4a7dd16d09d6f485347a2a379e464167533d61d8d24fbfR152) was crucial for importing to R.

I wrote some lines of demo code [here](https://gist.github.com/BerriJ/940419b119b97481038fb23bf7333db4).

eddelbuettel:
> and now it works.

That would be sweet!  I was tied up somewhere else (there was for once a bug in the RC for (Rcpp)Armadillo, plus other "stuff") but this is excellent new.  I think we can add ample unit tests and illustrations as this matures.

BerriJ:
I see that CI reports failing tests.

This affects`https://github.com/RcppCore/RcppArmadillo/blob/8c188323426604bc45191d66b4301b7e23859396/inst/tinytest/test_rcpparmadillo.R#L40`

At first glance, those are field/dim related and may need to be adjusted.

```
expect_equal( res[[4]][[1]], matrix(0:3, ncol = 2, nrow=2))#, msg = "field<int>" )
expect_equal( res[[4]][[2]], matrix(letters[1:4], ncol = 2, nrow=2))#, msg = "field<std::string>" ) 
```

eddelbuettel:
Yes, maybe comment the failing tests out 'for now' and add a big fat `# TODO: Revisit` or `# FIXME` comment?   It is quite possibly that these were set up 'to suit the buggy behavior'.

coatless:
Regarding the user test, I think that's because the list matrix is now being considered as an array instead of a matrix under the `class()` check. 

https://github.com/RcppCore/RcppArmadillo/runs/4097501712?check_suite_focus=true#step:5:110

```r
   call| expect_equal(res[[4]][[2]], matrix(letters[1:4], ncol = 2, nrow = 2))
   diff| Attributes: < Component "dim": Numeric: lengths (2, 3) differ >
   diff| target is matrix, current is array
  Error: 2 out of 259 tests failed

```

In R 4.0 higher, this object is considered to inherit both: `matrix` and `array` classes. 

Quickly checking through an example gives:

```r
A = matrix(1, ncol = 2, nrow = 3)
B = matrix(4, ncol = 5, nrow = 6)

listmatrix = matrix(list( ), 2, 2)

listmatrix[1, 1] = list(A)
listmatrix[1, 2] = list(B)
listmatrix[2, 1] = list(A)
listmatrix[2, 2] = list(B)

listmatrix
#          [,1]      [,2]      
# [1,] numeric,6 numeric,30
# [2,] numeric,6 numeric,30

typeof(listmatrix)
[1] "list"
class(listmatrix)
[1] "matrix" "array" 
```


```r
# Pass into C++
recovered_listmatrix = arma_field_exporter(listmatrix)

# Check list recovery
class(recovered_listmatrix)
[1] "array"

typeof(recovered_listmatrix)
[1] "list"
```

I think when I penned #263, I had expected the `class()` declaration to be in a `list` form. After a few years of reflection, I'm not sure pursuing an underlying class change to `list()` would be ideal. Perhaps a better approach that should be taken is a wrap of the `arma::field<>` with `Rcpp::List()`.

eddelbuettel:
> In R 4.0 higher, this object is considered to inherit both: matrix and array classes.

Correct but shouldn't it have bitten us then for all that time too?   (Maybe I am asleep, and tickling the bug needs our changes hence we only see it now...)  

In any event I think there is reasonably little use of `fields` in the wild (within R land) that we can probably clean up a little.  But we should get it right.

(Quick check)

Ok, proving myself wrong with one query.  GH mirror of CRAN has 100+ hits for `arma::field`.  So we can't be too radical.

eddelbuettel:
@BerriJ : Ok, we are all caught up at CRAN and with Armadillo upstream.  Can you please rebase to `master` ?

BerriJ:
> @BerriJ : Ok, we are all caught up at CRAN and with Armadillo upstream. Can you please rebase to `master` ?

I did. I also squashed my commits since the first 2 were just fiddling around with it. Now it's one commit with all of my changes.

> Ok, proving myself wrong with one query. GH mirror of CRAN has 100+ hits for arma::field. So we can't be too radical.

Maybe we can do a reverse dependency check and look how bad it is? I suspect that most people won't be affected by this change even if they use arma::fields. I think about the following:

- I think that most people use **arma::fields purely inside of C++** so they won't be affected at all.

- People who use **fields as input** arguments should be really rare.  Some may use 1d fields as input because it works as expected, but they should not be affected by this change at all. I don't think that anyone passes 2d or 3d lists to an arma::field. First, because the 2nd and 3rd dimensions will get lost, and second because the creation in R is not easy and I suspect that many people do not know about 2d and 3d lists in R.

- People who **return arma::field** to R will get the biggest hit. Currently, they do this using 1d and 2d fields only since 3d fields do not work. With the proposed changes in **RcppArmadilloWrap.h** they will get another dimension. So depending on if (and how) they use those output objects they may have to add an extra comma. But that should be all they have to do.

eddelbuettel:
> Maybe we can do a reverse dependency check and look how bad it is? 

Yup. I have the machinery for it, and fresh baselines from the three tests I ran for RcppArmadillo 0.10.7.3.0 its predecessor release candidates.  Stay tuned!

eddelbuettel:
So the test run and the result [has been comitted as usual](https://github.com/RcppCore/rcpp-logs/commit/e954193dd45a0a81165b9ba647e04c0db50424f3) with the summary in the table below.

That is a lot of shrapnel.  We cannot merge the PR as is.  I have been thinking about this while it was running and the results were accumulating towards this outcome:  could you possibly rejig the fix in the PR such that it is opt-in?  In other words set up a `#define` to switch in the two headers (or even switch within the two -- for `wrap` it seems easy to do in the file; for `as` we could do either way--in the file maybe simpler but less clear to read?).  

That way _you_ get to opt into new and corrected `fields` behaviour, and so can other if they choose but by default nothing changes.  We can then contact maintainer and suggest updates.  (I am currently doing this (since April !!) for Rcpp [here](https://github.com/RcppCore/Rcpp/issues/1158) and need to do one for RcppEigen to get to 3.4.0 so this won't be imminent but it is both doable and should be done.)

```
Test of RcppArmadillo 0.10.7.3.1 had 893 successes, 26 failures, and 8 skipped packages. 
Ran from 2021-11-06 08:48:31.83 to 2021-11-06 21:30:47.01 for 12.704 hours 
Average of 49.337 secs relative to 294.11 secs using 6 runners

Failed packages:  ADMM, AlphaSimR, Bayesrel, BGVAR, carat, cIRT, EMbC, FLightR, FMCCSD, grpsel, HDSpatialScan, higlasso, kernelPSI, meshed, netClust, NetMix, odpc, pdSpecEst, perccal, psgp, quanteda.textstats, rstpm2, simts, spamtree, T4transport, wv 

Skipped packages:  BMTME, Crossover, DataVisualizations, dynamichazard, growfunctions, joineRML, nlmixr, Rfast 

None still working

None still scheduled

Error summary:
               package missingPkg badInstall error fail warn note ok hasOtherIssue
 1:               ADMM                 FALSE     1    0    0    3 10         FALSE  likely new issue
 2:          AlphaSimR                 FALSE     0    0    0    6  8         FALSE  likely new issue
 3:           Bayesrel                 FALSE     0    0    0    0 14         FALSE  known earlier issue
 4:              BGVAR                 FALSE     4    0    0   10  0         FALSE  upstream, passes with new version
 5:              carat                 FALSE     0    0    0    4 10         FALSE  likely new issue
 6:               cIRT                 FALSE     0    0    0    1 13         FALSE  likely new issue
 7:               EMbC                 FALSE     0    0    0    0 14         FALSE  known earlier issue 
 8:            FLightR                 FALSE     0    0    0    7  7         FALSE  known earlier issue
 9:             FMCCSD                 FALSE     0    0    0    0 14         FALSE  likely new issue
10:             grpsel                 FALSE     0    0    0    1 13         FALSE  likely new issue
11:      HDSpatialScan                  TRUE     0    0    0    0 14         FALSE  known earlier issue
12:           higlasso                 FALSE     0    0    0    0 14         FALSE  likely new issue
13:          kernelPSI                 FALSE     0    0    0    0 14         FALSE  likely new issue
14:             meshed                 FALSE     1    0    0    6  7         FALSE  likely new issue
15:           netClust                 FALSE     0    0    0    3 11         FALSE  likely new issue
16:             NetMix                 FALSE     0    0    0    0 14         FALSE  likely new issue
17:               odpc                 FALSE     0    0    0    0 14         FALSE  likely new issue
18:          pdSpecEst                 FALSE     0    0    0   13  1         FALSE  likely new issue
19:            perccal                 FALSE     0    0    0    2 12         FALSE  likely new issue
20:               psgp                  TRUE     0    0    0    4 10         FALSE  known earlier issue
21: quanteda.textstats                 FALSE    10    0    0    2  2         FALSE  known earlier issue
22:             rstpm2                 FALSE     0    0    0    6  8         FALSE  known earlier issue
23:              simts                 FALSE     0    0    0    7  7         FALSE  likely new issue
24:           spamtree                 FALSE     1    0    0    4  9         FALSE  likely new issue
25:        T4transport                 FALSE     0    0    0    0 14         FALSE  likely new issue
26:                 wv                 FALSE     0    0    0    7  7         FALSE  likely new issue
               package missingPkg badInstall error fail warn note ok hasOtherIssue
```

eddelbuettel (approved):
Nice work also to re-enable the tests/

BerriJ:
> Nice work also to re-enable the tests/

Thank you. So I simply defined 2 variables:
// #define RCPP_ARMADILLO_FIX_FieldExporter
// #define RCPP_ARMADILLO_FIX_FieldImporter
to control the behavior in the 2 files respectively.

However, I could not test it. I tried simply running those functions:

test.cpp:

```
#define RCPP_ARMADILLO_FIX_FieldExporter
#define RCPP_ARMADILLO_FIX_FieldImporter

#include <RcppArmadillo.h>
using namespace arma;

// [[Rcpp::export]]
arma::field<arma::mat> fieldtest1()
{
    arma::field<arma::mat> out(5, 5);

    return out;
}

// [[Rcpp::export]]
arma::field<arma::mat> fieldtest2(arma::field<arma::mat> x)
{

    arma::cout << x.n_rows << arma::endl;
    arma::cout << x.n_cols << arma::endl;
    arma::cout << x.n_slices << arma::endl;

    return x;
}
```

after executing devtools::load_all(). Probably my setup does not catch that `#include <RcppArmadillo.h>` late enough.
Putting those defined statements in the 2 files directly worked fine. Maybe you have an idea how you would test that?

I hope I got it. Ci looks good at least :+1: .

Also, it would be very interesting to make 2 additional reverse dep checks to see which change is more problematic.

Thanks so far, Dirk.

eddelbuettel:
>  Maybe you have an idea how you would test that?

We should be able to add something.  One approach is simply ... another cpp source file _with_ the `#define`s set as you have above.  I should be able to probe (and we can add that post merge).

New reverse depends run started and now past the first former failure so fingers crossed ...

inst/include/RcppArmadilloWrap.h
```diff
@@ -149,7 +149,11 @@ namespace Rcpp{
     template <typename T>
     SEXP wrap( const arma::field<T>& data){
         RObject x = wrap( RcppArmadillo::FieldImporter<T>( data ) ) ;
-        x.attr("dim" ) = Dimension( data.n_rows, data.n_cols ) ;
+        # if defined(RCPP_ARMADILLO_FIX_FieldImporter)
+            x.attr("dim" ) = Dimension( data.n_rows, data.n_cols, data.n_slices ) ;
```
coatless, line 153: Shouldn't this be further conditioned based on the dimension attributes of `data` similar to the as call? , e.g. 

```
                if (data.n_elem == 1)
                {
                   x.attr("dim" ) = Dimension(data.n_rows);
                }
                else if (data.n_elem == 2)
                {
                     x.attr("dim" ) = Dimension( data.n_rows, data.n_cols);
                }
                else if (data.n_elem == 3)
                {
                     x.attr("dim" ) = Dimension( data.n_rows, data.n_cols, data.n_slices);
                }
```

If retained as-is, this would force all fields to be an array with 3 dimensions in _R_.

coatless:
> So the test run and the result [has been comitted as usual](https://github.com/RcppCore/rcpp-logs/commit/e954193dd45a0a81165b9ba647e04c0db50424f3) with the summary in the table below.

Odd to see `cIRT` appear under that header. Could this be picking up the new macos m1 check notes on CRAN?

eddelbuettel:
Absolutely not. The rev.dep box is a plain standard Linux machine with an Intel CPU.  Always has been.  What happens in other run on other machines under other OSs is, well, something else. 

It is a regular error likely caused by the `field` changes:

```
Running examples in ‘cIRT-Ex.R’ failed
The error most likely occurred in:

> ### Name: direct_sum
> ### Title: Direct Sum of Matrices
> ### Aliases: direct_sum
> 
> ### ** Examples
> 
> 
> x = list(matrix(0, nrow = 5, ncol = 3),
+          matrix(1, nrow = 5, ncol = 3))
> direct_sum(x)
Error in direct_sum(x) : 
  Not compatible with requested type: [type=NULL; target=integer].
Execution halted
* checking for unstated dependencies in vignettes ... OK
* checking package vignettes in ‘inst/doc’ ... OK
* checking running R code from vignettes ... SKIPPED
* checking re-building of vignette outputs ... SKIPPED
* DONE
Status: 1 ERROR, 1 NOTE
```

inst/include/RcppArmadilloWrap.h
eddelbuettel: Nice catch, thanks for that.

eddelbuettel:
Last rev.dep was of course again very clean, see [its commit](https://github.com/RcppCore/rcpp-logs/commit/5b2baa743b09e7b475002be2ccbd3270f95b864b) 

So I think we're ready, modulo maybe reflecting on the [previous comment](https://github.com/RcppCore/RcppArmadillo/pull/352#discussion_r744304025).  Do you want to fold that in?

BerriJ:
I would propose to merge as is. The main functionality of exporting and importing fields works fine and always returning fields with 3 dimensions is something most people can easily live with I think.

One could think about adding an additional flag for dropping dimensions (something like drop=TRUE in R). But I would propose to postpone this for now since I'm quite busy.

eddelbuettel:
Hi @BerriJ sorry for this taking a moment but we may as well get it right.  I just committed (to master) a file `local/armafield.cpp` which creates a few `field` objects, prints them to stdout and tries to return them to R.  The last one a 3-d field, and this still fails (under `master` and under your PR), and I think it shouldn't.

coatless:
@eddelbuettel agreed. The proposed change is still a bit problematic. There shouldn't be a reason for why [`cIRT::direct_sum()`](https://github.com/tmsalab/cIRT/blob/50e6ebf740f5b65f4855a5d4fb8b35b343f05499/src/armadillo_manipulations.cpp#L33-L74) is failing as that routine interacts with a field object using only index notation.

BerriJ:
> Hi @BerriJ sorry for this taking a moment but we may as well get it right. I just committed (to master) a file `local/armafield.cpp` which creates a few `field` objects, prints them to stdout and tries to return them to R. The last one a 3-d field, and this still fails (under `master` and under your PR), and I think it shouldn't.

Did you maybe miss executing a git push? I can't find the commit you refer to. Looked [here](https://github.com/RcppCore/RcppArmadillo/commits/master).

eddelbuettel:
I did indeed, sorry for that and thanks for the heads-up.  Sometimes I think the 'sequence' is muscle memory in `magit` but every now and it isn't.

BerriJ:
I'm pretty shure I got cIRT::direct_sum() working now. The problem was that it crashed at `arma::ivec dims = data.attr("dim");` for List inputs that had no `dim() = NULL`. I just added a check > `arma::ivec dims = data.attr("dim");` If `true` it will set the field size according to dims, if `false` it will just make a 1D field with length of the list (basically the old behavior).

I still have to look into your file @eddelbuettel .

BerriJ:
Sorry for adding another commit. I realized that I had left `#define RCPP_ARMADILLO_FIX_FieldExporter` in RcppArmadilloAs.h.
The code in `local/armafield.cpp` now runs exactly as I expect it to.

So as far as I see we are done now? Do you want to run the rev dep checks another time?

eddelbuettel:
Yes I think we're there, or very close.  

The #ifdef protection should mean to harm over at the CRAN side and those who want it (you ;-) ) can opt into it now.  We can then slowly and patiently email those with failing packages and suggest they turn the `#define` and adjust their code so that "eventually" we can turn it on by default.

eddelbuettel:
Oh, the one thing I forgot: as I made one change to `master` adding the fragment of a to-be-refined test script, could you be please rebase once more?

BerriJ:
> Oh, the one thing I forgot: as I made one change to `master` adding the fragment of a to-be-refined test script, could you be please rebase once more?

I did, no problem. :-)

eddelbuettel:
Thanks for your patience and diligence on this!

PR #381, opened 2022-06-12
RcppArmadillo 0.11.2.0.0 with Armadillo 11 2

eddelbuettel:
This version is ready and has been submitted to CRAN but awaiting a review on what is suspected to be a spurious fail.

Reverse-dependency checks passed with flying colours, see https://github.com/RcppCore/rcpp-logs/commit/aa6dad36c11f4287934b306c3b2f0c4156ceebf7


conradsnicta:
For posterity and transparency, the fail is in the KODAMA package (see [CRAN package](https://cran.r-project.org/web/packages/KODAMA/) and [package mirror on GH](https://github.com/cran/KODAMA)).

The KODAMA package appears to have at least two bugs in the [src/prova.cpp](https://github.com/cran/KODAMA/blob/585e14cae68dbaebd16f9e581971873bd7c5f5ff/src/prova.cpp) file, on the following lines:
* https://github.com/cran/KODAMA/blob/585e14cae68dbaebd16f9e581971873bd7c5f5ff/src/prova.cpp#L608-L609
* https://github.com/cran/KODAMA/blob/585e14cae68dbaebd16f9e581971873bd7c5f5ff/src/prova.cpp#L858-L859

In both cases the bug appears in:

    svd_econ(svd_U,svd_s,svd_V,S,"left");
    rr=svd_U.col( 0 );

The bug is in `svd_U.col(0)`.  This fails because the `svd_U` matrix is empty, and hence `svd_U.col(0)` attempts to access a non-existent column.

The `svd_U` matrix is empty because the `svd_econ()` function failed.  Since Armadillo 11.0, the [`svd_econ()`](http://arma.sourceforge.net/docs.html#svd_econ) function will reject matrices with non-finite elements (NaN or Inf).  At failure, the `svd_econ()` function returns a bool set to false, and resets all output matrices.  In this case, the input matrix `S` is full of NaNs.

One possible solution is to check the success of the `svd_econ()` function before proceeding. For example:

    bool ok = svd_econ(svd_U,svd_s,svd_V,S,"left");
    if(ok == false)  { /* handle failure */ }

It may also be useful to ensure that the matrix fed to `svd_econ()` doesn't have any NaNs.  The [`.is_finite()`](http://arma.sourceforge.net/docs.html#is_finite) member function can be used here.  For example:  `if(S.is_finite() == false)  { ... }`

conradsnicta:
No worries.  I'll contact the authors directly.

eddelbuettel:
Thank you!   The CRAN thing holding the release up did not look like a test fail, and as I said, I still haven't heard so ... let's sit tight for a moment.

eddelbuettel:
This is being held up at CRAN with an issue I _suspect_ to be spurious but we will not know for sure until we hear from CRAN.

Either way, it revealed a possible issue an upstream package due to subtle but important changes in Armadillo, and @conradsnicta followed up with a very nice email to the upstream author.

eddelbuettel:
Merging this now as I am tired of waiting on CRAN you have in four+ days incl two full business days not have had the time of day to reply to me.  Which is rather sad but such is life.

PR #388, opened 2022-08-17
simplification

conradsnicta:
arma::dot() is simpler and more intuitive to use than std::inner_product().  

arma::dot() is also faster than std::inner_product(). Armadillo maps arma::dot() to BLAS ddot(), which is accelerated by OpenBLAS (or MKL).

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ Rcpp::List fastLm(const arma::mat& X, const arma::colvec& y) {
     arma::colvec res  = y - X*coef;           // residuals
 
     // std.errors of coefficients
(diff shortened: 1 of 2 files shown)


coatless:
@conradsnicta could you also modify the routine in `src/fastLm.cpp`:

https://github.com/RcppCore/RcppArmadillo/blob/3b346f42b67ecf86959c4903c4fb211ab65eea72/src/fastLm.cpp#L33

eddelbuettel:
Sure, we'll get to that. I also need to add a ChangeLog entry.  Currently waiting (two days now) on CRAN to ack that the one fail in one test was in fact spurious as I am unable to reproduce on either an Ubuntu or Debian system ... and then we'd have a new version with the changes requested by the `Matrix` maintainers.

PR #395, opened 2022-10-12
Reorganise internal headers

eddelbuettel:
Attempting the clean up inside inst/include/ and moving clearly internal headers further afar.


inst/include/RcppArmadillo/Lighter
```diff
@@ -0,0 +1,44 @@
+// Emacs, please make this  -*- mode: C++; -*-
```
coatless, line 1: Missing `.h` from `Lighter`.

eddelbuettel: Thanks for noticing but that is actually on purpose.  See Rcpp:

```sh
edd@rob:~/git/rcpp(bugfix/windows_path)$ ls -l inst/include/Rcpp/{Rcpp,Light,Lighter,Lightest}
-rw-rw-r-- 1 edd edd  932 Dec 10  2021 inst/include/Rcpp/Light
-rw-rw-r-- 1 edd edd  997 Dec 10  2021 inst/include/Rcpp/Lighter
-rw-rw-r-- 1 edd edd 1029 Dec 10  2021 inst/include/Rcpp/Lightest
-rw-rw-r-- 1 edd edd  840 Dec 10  2021 inst/include/Rcpp/Rcpp
edd@rob:~/git/rcpp(bugfix/windows_path)$ 
```

But the move didn't quite work out in rev.dep.  See https://github.com/RcppCore/rcpp-logs/blob/master/results/rcpparmadillo/RcppArmadillo-20221012.txt -- `mlpack` is one of the bad ones. So I will move the RNG header as well as *Config.h and *Forward.h back.

eddelbuettel:
Per extensive and repeated reverse-depends checks, this 'does no harm' but offers some new choices.  So good to go.

Issue #406, opened 2023-02-10
package compilation failure


mansillo:
This is perhaps an esoteric approach to installing the RcppArmadillo package update but here goes. I routinely update my R packages in RStudio with the check for updates function. Every single time that the RcppArmadillo package appears in that list it fails to compile. I restart the session, restart R, terminate R, restart the computer etc and various iterations of these and the package stubbornly does not compile. Then typically a day or so later I have found in the past without an explanation I can identify the package updates. This drives me utterly mad.

I run the code ... 
`install.packages("RcppArmadillo")`

If anyone can identify why this occurs locally, I'd be interested to know:
 >  installing *source* package ‘RcppArmadillo’ ...
 > ** package ‘RcppArmadillo’ successfully unpacked and MD5 sums checked
 > ** using staged installation
 > checking whether the C++ compiler works... yes
 > checking for C++ compiler default output file name... a.out
 > checking for suffix of executables... 
 > checking whether we are cross compiling... no
 > checking for suffix of object files... o
 > checking whether the compiler supports GNU C++... yes
 > checking whether clang++ -mmacosx-version-min=10.13 -std=gnu++14 accepts -g... yes
 > checking for clang++ -mmacosx-version-min=10.13 -std=gnu++14 option to enable C++11 features... none needed
 > checking how to run the C++ preprocessor... clang++ -mmacosx-version-min=10.13 -std=gnu++14 -E
 > checking whether the compiler supports GNU C++... (cached) yes
 > checking whether clang++ -mmacosx-version-min=10.13 -std=gnu++14 accepts -g... (cached) yes
 > checking for clang++ -mmacosx-version-min=10.13 -std=gnu++14 option to enable C++11 features... (cached) none needed
 > checking whether we have a suitable tempdir... /var/folders/gm/lg9j9k791ms06zsbp6r9zk500000gn/T
 > checking whether R CMD SHLIB can already compile programs using OpenMP... yes
 > checking LAPACK_LIBS... R-supplied partial LAPACK found
 > configure: WARNING: Some complex-valued LAPACK functions may not be available
 > configure: creating ./config.status
 > config.status: creating inst/include/RcppArmadillo/config/RcppArmadilloConfigGenerated.h
 > config.status: creating src/Makevars
 > ** libs
[...]

PR #407, opened 2023-02-11
set includes in PKG_CPPFLAGS

kevinushey:
Fixes the underlying issue in https://github.com/RcppCore/RcppArmadillo/issues/406. With this, the compiler invocation when building on macOS is (for example):

```
clang++ -mmacosx-version-min=10.13 -std=gnu++14 -I"/Library/Frameworks/R.framework/Resources/include" -DNDEBUG -I../inst/include -I'/Users/kevin/Library/R/x86_64/4.2/library/Rcpp/include' -I/usr/local/include   -fPIC  -Wall -g -O2  -c RcppArmadillo.cpp -o RcppArmadillo.o
< ... >
```

And because `-I../inst/include` is now coming before `-I/usr/local/include`, we should now be able to avoid issues where a system installation of Armadillo in `/usr/local/include` takes precedence over the bundled version.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+2023-02-11  Kevin Ushey  <kevinushey@gmail.com>
+
+	* src/Makevars.in: Use PKG_CPPFLAGS for RcppArmadillo includes
+
 2023-02-09  Dirk Eddelbuettel  <edd@debian.org>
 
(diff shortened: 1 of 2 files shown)


eddelbuettel (approved):
That's nice and should not create headaches in other uses / OSs yet protect against the case we saw here.

PR #417, opened 2023-06-04
Accomodate upcoming change in package Matrix (fixes #415)

### LINKED ISSUE #415: Update indMatrix->arma::SpMat and pMatrix->arma::SpMat for Matrix 1.5-5
In **Matrix** 1.5-5, class `indMatrix` will gain a slot `margin` with value `1L` or `2L`.  If `x` is an `indMatrix` and `x@margin = 1L` (the "prototype"), then the remaining slots of `x` are interpreted as in older versions of **Matrix**: `x` has dimensions `x@Dim` with names `x@Dimnames` and is zero except for ones in positions `cbind(i = seq_along(x@perm), j = x@perm)`.  If `x@margin = 2L`, then the ones occur in positions `cbind(i = x@perm, j = seq_along(x@perm))` instead.

In general, we try to avoid making such changes to class definitions, as they can invalidate serialized objects.  But this particular change (a very useful generalization) is long overdue and, perhaps amazingly, **RcppArmadillo** is the only broken reverse dependency.

AFAICT, it should suffice to update the `indMatrix` and `pMatrix` branches in `inst/include/RcppArmadillo/interface/RcppArmadilloAs.h` to behave conditional on `margin`, checking first for the _absence_ of a `margin` attribute (implying that an old **Matrix** is installed) and in that case behaving as if `margin = 1L`.

We have not discussed a release date for 1.5-5, so this is just an FYI (for now).  In the mean time, I can advise on a patch if necessary.

Note that R-Forge is currently hosting an out-of-date **Matrix** tarball.  Until that is fixed, you can use the R-universe tarball for testing (or install from sources if you have Subversion).

jaganmn:
Finally got around to working on this.  A few comments:

* Perhaps verify that I haven't introduced whitespace issues. I noticed that my Emacs inserted tabs where the file otherwise had none.  Well, I fixed whatever I saw ...
* Unrelated to the issue, I replaced 5 lines of code repeated throughout the file with a macro.  It was too upsetting ...
* I noticed that in several places you have `ncol + 1`, where `ncol` is an `int`.  Note that `ncol` may take the value `INT_MAX` and in that case `ncol + 1` will overflow.  The patch does not attempt to address this.
* I've used `R_alloc` somewhere, but maybe that is not idiomatic ... I needed to do `a[i]++` and did not feel like checking whether that was legal for `a` of type `IntegerVector` rather than `int *` ...

### DIFF
--- a/inst/include/RcppArmadillo/interface/RcppArmadilloAs.h
+++ b/inst/include/RcppArmadillo/interface/RcppArmadilloAs.h
@@ -383,10 +383,9 @@ namespace traits {
 
                 if (!mat.hasSlot("margin") ||
                     as<IntegerVector>(mat.slot("margin"))[0] == 1) {
-                    for (int jj = 0; jj < ncol; ++jj)
-                        i[perm[jj] - 1] = jj;
                     for (int jj = 0; jj < ncol; ++jj) {
                         p[jj] = jj;
+                        i[perm[jj] - 1] = jj;
                         x[jj] = 1;
                     }
                 } else {
--- a/inst/include/RcppArmadillo/interface/RcppArmadilloAs.h
+++ b/inst/include/RcppArmadillo/interface/RcppArmadilloAs.h
@@ -123,21 +123,30 @@ namespace traits {
 
             // Get the type of sparse matrix
             std::string type = Rcpp::as<std::string>(mat.slot("class"));
+
             if (type == "dgCMatrix" || mat.is("dgCMatrix")) {
                 IntegerVector i = mat.slot("i");
                 IntegerVector p = mat.slot("p");
                 Vector<RTYPE> x = mat.slot("x");
 
-                // Making space for the elements
-                res.mem_resize(static_cast<unsigned>(x.size()));
+#define DO_RESULT                                                       \
+                do {                                                    \
+                    /* Allocate: */                                     \
+                    res.mem_resize(static_cast<unsigned>(x.size()));    \
+                                                                        \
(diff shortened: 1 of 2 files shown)


eddelbuettel:
Thanks for the follow-up!  I had getting back to you on the TODO list has I was left a little bewildered and lost about what you wanted me to do.  Will peruse the PR -- thanks for sending it in!

inst/include/RcppArmadillo/interface/RcppArmadilloAs.h
```diff
@@ -123,38 +123,38 @@ namespace traits {
...
+                    /* Copy: */                                         \
+                    std::copy(i.begin(), i.end(),                       \
+                              arma::access::rwp(res.row_indices));      \
+                    std::copy(p.begin(), p.end(),                       \
+                              arma::access::rwp(res.col_ptrs));         \
+                    std::copy(x.begin(), x.end(),                       \
+                              arma::access::rwp(res.values));           \
+                } while (0)
+
+                DO_RESULT;
```
eddelbuettel, line 149: Old school :)   Maybe we have too much C around R to not let go of macros.

eddelbuettel (commented):
This looks pretty good, and is actually not code I have been looking at much (all this was part of a nice Google Summer of Code project).  

Could you add an entry to the ChangeLog in standard Emacs ChangeLog mode (ie two spaces between date and name, and between name and email?

eddelbuettel (approved):
Thanks!

PR #449, opened 2024-08-01
Swap order of the two bibentry instances

conradsnicta:
@eddelbuettel  The peer-reviewed article should be cited before the manual.  Detailed reasoning given in email.

### DIFF
--- a/inst/CITATION
+++ b/inst/CITATION
@@ -1,6 +1,3 @@
-bibentry("Manual",
-         other = unlist(citation(auto = meta), recursive = FALSE))
-
 bibentry("Article",
          title = "RcppArmadillo: Accelerating R with high-performance C++ linear algebra",
          author = c(person("Dirk", "Eddelbuettel",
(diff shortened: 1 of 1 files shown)


eddelbuettel:
It would have helped if you had let have an hour or two to respond to your email (which arrived in the middle of my night).

eddelbuettel:
As per my email reply, we have more work to do on package metadata also influencing citation.  We can do this here, or in separare PRs.  I am mostly indifferent but it may make sense to hash it all out here.  Thoughts?

conradsnicta:
separate PRs are preferred

eddelbuettel:
I lean towards doing it here in one PR dealing with 'citation() output'.  But I will mull it over some more.

conradsnicta:
I'm time constrained, so the current PR is the best that I can do

eddelbuettel:
See email. For the related other part I need input from you, or else need to proceed with entries you may find suboptimal.  Your call.

This is not urgent.  CRAN is now closed for two weeks, and we aim to only upload "once every 1 - 2 months" (quoting their guidelines) to not overwhelm CRAN.

PR #460, opened 2024-12-03
Sync with Armadillo 14.2.x branch

eddelbuettel:
This is chiefly motivated by a friendly nudge from CRAN to get (Rcpp)Armadillo to build under the current gcc 15 pre-release which this update accomplishes.

_Edit:_  And it morphed into a proper new release following an Armadillo release.  Now also at CRAN.


inst/include/armadillo_bits/arma_version.hpp
```diff
@@ -23,7 +23,7 @@
 
 #define ARMA_VERSION_MAJOR 14
 #define ARMA_VERSION_MINOR 2
-#define ARMA_VERSION_PATCH 1
+#define ARMA_VERSION_PATCH 2
```
conradsnicta, line 26: @eddelbuettel Not quite yet.  I've reverted this change on GitLab.

I've added more commits to armadillo branch 14.2.x. This is to ensure consistency and prevent further potential regressions under gcc 15.

eddelbuettel: @conradsnicta As you may have noticed I have not merged yet.

conradsnicta:
For posterity, the regression in GCC 15 has been reported: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=117993

PR #466, opened 2025-04-25
RcppArmadillo 14.4.2-1 with Armadillo 14.4.2

eddelbuettel:


eddelbuettel:
And by now we have 

> Thanks, on its way to CRAN.

PR #505, opened 2026-04-16
Armadillo 15.2.5

eddelbuettel:


eddelbuettel:
We do have one regression, and an [issue](https://github.com/bsvars/bsvarSIGNs/issues/75) has been filed.

